PR #1, opened 2019-03-26
Refactor codebase

watadarkstar:
The goal of this PR is to make the codebase a little bit more scalable / maintainable before I start getting into the UI / UX.

- Move components into a common folder
- Adds a pages folder in components
- Creates a Home page component
- Refactors `App.js` into a Home component (we will break this out more later)
- Adds eslint
- Adds prettier
- Format codebase using prettier
- Move styles into their own files


pvinis:
👍

PR #7, opened 2019-05-31
Upgrade guide

lucasbento:
# Summary

This PR introduces a _very initial_ upgrade guide for React Native.

## Test Plan

1. Clone the project
1. Run `yarn install`
1. Run `yarn start`
1. Open `http://localhost:3000`
1. Select a `0.58.x` version
1. Click the `Show me how to update` button

### TODOs

- [x]  Fix the width of the select inputs
- [x]  Hide the "Show me how to update" button after it's clicked the first time
- [x]  Filter out the versions on the "to version" select after the "from version" is set
- [x]  Show a *major* version bump when rendering the first time
- [x]  Add filter to toggle the visibility of `Added`, `Changes`, `Breaking Changes` & `Troubleshooting`
- [x]  Add a way to collapse the diff
- [x]  Highlight changes like GitHub does
- [x]  Handle deleted files
- [x]  Show a type for each change (add, delete, modify and rename)
- [x]  Collapse a few diffs by default (currently collapsing deleted files & diff with more than 5 hunks)
- [x]  Hide `RnDiffApp/` prefix
- [x]  Always show `package.json` at the top (thanks @matt-oakes :wink:)
- [x]  Only change the diff after clicking the button

---

image

---

image

---

image


src/components/versions/0.59.0.js
```diff
@@ -0,0 +1,42 @@
+import React, { Fragment } from 'react'
```
lucasbento: @pvinis, @kelset: this is a file that will be created for each version, I would appreciate feedback on this very much.

pvinis:
I'll be checking it later today 💪💪

ronal2do:
awesome work

src/components/versions/0.59.0.js
pvinis: I would only change the name, to `releases` instead of `versions`, to align with the way RN calls them.

lucasbento: Good call.

lucasbento:
To show the instructions you must select any `0.58.x` version and `0.59.x` or `0.60.x`, the reason for this is that only the `0.59.x` release exists there.

---

> I kinda hate the styling (ant?). Everything feels slow, and I noticed that the dropdowns and buttons act on click up, and not click down.

Why do you feel it's slow? is it because of the click up/down thing? I'll check if I can find any way to trigger them on click up but I'm not sure if that's the thing that's giving you the slow feeling. Also, I chose a UI lib to speed it up, I would love to hear other opinions on libraries that we could use.

> Good that we changed it back to one file per component, I didn't love the two files, but it was ok. In three PRs we have changed from react-native-web, to react-native-web with separate styles, to styled-components :D.

I'm just trying stuff and seeing how they fit, first time using `styled-components` here, nothing is set on stone so we can definitely discuss those approaches :)

> I don't love having .vscode stuff in repos.

I'll remove it 👍

src/components/pages/Home.js
```diff
@@ -43,7 +43,7 @@ const Home = () => {
   const [toVersion, setToVersion] = useState('')
   const [packageManager, setPackageManager] = useState(packageManagers[0])
   const [showUpdateGuide, setShowUpdateGuide] = useState(false)
-  const [filters, setFilters] = useState(Object.keys(INSTRUCTION_CATEGORIES))
+  const [filters, setFilters] = useState(Object.values(INSTRUCTION_CATEGORIES))
```
lucasbento: @pvinis: the reason it wasn't working for you was this, it was filtering with incorrect values, should work now :)

src/components/versions/0.59.0.js
kelset: If I understand, this would be basically something that needs to be manually created per each version (ex. 0.59.0, 0.59.1, etc)?

Is there a quick way we could parse the git diff for, let's say, a v1.0, and then migrate to have a custom file in a second version?

kelset: (And in this gitdiff version, the checklist could be at file level)

lucasbento: > (And in this gitdiff version, the checklist could be at file level)

What do you mean by this?

A diff wouldn't be much effective, the idea of having the upgrade guide is to have more description about what to change, how to do it and perhaps the reason, correct me if I'm wrong though.

I was thinking that we can do it manually in the beginning and get something like parsing the `CHANGELOG.md` to create these files and supply more context later on.

pvinis: Yea, parsing the changelog would be a way.

pvinis:
I'll try it again.

Yea, the ant is feeling slow to me because of the click up/down. (is that where the name ant comes from? 🤣🤣😬😛)

I guess styled-components is a nice way to go for now. Would also be cool to keep using rn-web even with the styled-components? Kinda dogfooding, it's nice, right?

src/components/versions/0.59.0.js
kelset: So, I mean, it could have a checkbox near each one of the file names, not too dissimilar to this https://github.com/react-native-community/rn-diff-purge/compare/release/0.59.8..release/0.60.0-rc.0

The reason why I'm trying to think of a "less manual" implementation is that it would be allow to backport/use this webapp also for older versions.

I think that we should have something that is automatic and then, if people like it and use it, add on top some custom/manual things.

Re: parsing the changelog. I'm not sure that would work, meaning that the changelog will have all the changes so it feels that we'd have to do twice the work, parse first, then group the changes, then show them. Not sure, I wish there was an easier way. Maybe starting from the patch file? https://raw.githubusercontent.com/react-native-community/rn-diff-purge/diffs/diffs/0.59.8..0.60.0-rc.0.diff

lucasbento:
@pvinis: I was thinking about that and at some point I believe it just adds more complexity to the project, the main pro is that RN devs will be familiar with the syntax but I believe that most of them came from web so they probably know their way around it.

pvinis:
Woo super good!

- I still think the ant thing on click up is annoying.
- When there are no instructions, we should still show the `install react` and `install rn` I think, right?
- Maybe we can use the `filter` icon or the `check-square` one for the top right? Unless we might add more stuff there, so we can keep the `settings` one.

lucasbento:
@pvinis & @kelset: updated the PR to match what we discussed, now it shows the diff between `react-native` versions based on `rn-diff-purge`, I still have a few TODOs left (being tracked [here](https://github.com/react-native-community/purge-web/pull/7#issue-283945441)) that I'll be working on.

@pvinis about your feedback, I'll keep that in mind for a v2 as we won't be doing those instructions in this PR, I still need to clean up the files that are not being used though. Also, if you could suggest another UI library that you prefer we can talk about it :)

kelset:
Hey Lucas - I'm just checking it out locally and I have to say - it looks SO GOOD.

I'm honestly so happy that this is coming together, and I'm thankful you folks have been working on this 🙇‍♂️

Couple of things:

1) I think we should merge this PR, and do the subsequent features as smaller separate PRs. This way, we can open the door to multiple people doing multiple things at the same time.

2) Let's move the roadmap to an issue, so we will not lose the ideas once the PR is merged -> https://github.com/react-native-community/purge-web/issues/10

3) I'm not that bothered about the ant design, I think that if we want to revisit that in the near future it's fine but it shouldn't be a blocker.

lucasbento:
I discussed the ant design issue with @pvinis today and we will check how it will work without it as this uses only very simple stuff from UI libs.

Completely agree that we can ship it the way it is right now.

lucasbento:
As we talked, I believe this is good to go and we will keep working on different issues in the next PRs, @pvinis please give us your blessing so we can ship this 🚀

Issue #10, opened 2019-06-07
Roadmap


kelset:
With this issue I want to line up all the things that we are planning to do in a bullet point manners, so that we can sort of prioritise them and discuss/add/remove new ones.

### Version 1

*This version should be literally the barebone version, so that we can put it in front of the developers and have them start using it & provide feedback.*

- [x] Improve README with useful information
- [x] Add a way to conclude diffs as when they are done (PR at #21) - @lucasbento
- [x] Add useful links section (pointing to a blog post or something like that) (PR at #24)
- [x] Handle binaries (add a way for the user to download them) #22 (PR at #24)

### Version 2

*This one should be the "super powered" version of 1, with more advanced features & cool extras.*

- [x] Add comments on the diff (will be manually created by us, PR at #18) - @lucasbento
- [ ] review the ant design / maybe change it?
- [ ] add the versions in the url (shareability <3) #9

pvinis:
I would move the binaries to v2, but I agree with everything else.

For now we can just get the binaries by doing `init`. Not amazing, but easier than figuring out how to put these somewhere, hosting, linking etc.

lucasbento:
I believe we can force the binaries to be downloaded from GitHub, just need a few mins to investigate that.

lucasbento:
@kelset I added the README improvement to v1, would you be able to work on that?

kelset:
Yeah, I just checked the PRs and I think that once #24 & #18 are merged (optional to have #20 too) we can consider the v1 ready - and I can work on the Readme :3 

I was thinking on using this as inspiration: https://github.com/react-native-community/template/issues/4

lucasbento:
@kelset: awesome!

kelset:
Let's close this in favour of the project :D

PR #18, opened 2019-06-08
Show comments in the diff

lucasbento:
# Summary

This PR adds the ability to show comments in the diff, the comments are provided by a file under `./src/releases`. Related to #10.

## Test Plan

Run the app locally and select `0.60.0.rc-0` as the destination version.

---

image

---

image


pvinis:
🤣 I looooove those comments! I don't even need to talk about the second one!

kelset:
So the only blocker here for now is the copy for the comments?

lucasbento:
@kelset: I'm waiting for @pvinis to approve the code so we can merge it, before merging it I'll remove the current comments, right now they are here just to illustrate how to add them.

kelset:
Gotcha - I guess we want to have that in the readme right? Because these comments will live in this repo

pvinis:
I guess they will live here yes. But I didn't get what you mean with put it in the readme?

kelset:
oh sorry 😅 yeah I meant, like a section of the readme explaining how comments inline can be added

kelset:
my only concern is, what will happen if we have let's say a jump from 57 to 59, and the comment that we want to show was something that would be in 58?

lucasbento:
@kelset: it will show anyway unless `0.59` has a comment on the same line.

lucasbento:
@kelset: about adding docs for comments I created #28 to keep track of it.

PR #21, opened 2019-06-10
Add button to complete diff

lucasbento:
# Summary

This PR adds a button to complete a file diff, related to #10.

Currently, if you tap the button it will automatically collapse the diff file.

I'm not very much fan of the "check" button so I'm definitely open for opinions.

cc @kelset.

---

showing button to complete diff

---

image

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -61,6 +61,9 @@ const Diff = ({
   newPath,
   type,
   hunks,
+  diffKey,
+  isDiffCompleted,
+  onCompleteDiff,
   selectedChanges,
   onToggleChangeSelection
 }) => {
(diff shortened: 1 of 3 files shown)


lucasbento:
IMO currently the "complete" diff button is not very useful when compared to the collapse one, I'm thinking about adding a text or a floating thingy on the bottom right with `2/10` to indicate how many files were done out of 10, then it will make more sense.

If we prefer to ship this without the _counter_ then I would say we can hide the collapse button for now.

kelset:
tbh it looks so good - wanna merge it? 🤗

lucasbento:
Let's wait for @pvinis to review it

lucasbento:
@pvinis I'll create an issue for this!

PR #32, opened 2019-06-13
Add versions in URL to improve shareability

### LINKED ISSUE #9: add the versions in the url (shareability <3)
# Feature Request

We should have links like `gh.com/purge?from=0.30&to=0.39` to allow for linking and history keeping.

## Why it is needed

For sharability.

lucasbento:
# Summary

This PR fixes #9 by introducing the ability to specify versions in the URL.

If you provide versions in the URL the application will automatically show the diff, this PR also adds a functionality to mirror the versions that you select in the URL itself.

The following URLs are currently supported and will automatically show diff from `0.58.6` to `0.59.9`:

- `/?from=0.58.6&to=0.59.9` 
- `/0.58.6..0.59.9`

The following URLs are "handled":

1. `/?from=0.58.6`: this will specify `0.58.6` as the _current_ version and will set the target version automatically (latest, currently `0.60.0-rc.1`)
1. `/?to=0.59.9`: this will set the target version to `0.59.9` and **assume** the _current_ version as a major bump down (in this case it will be `0.58.6`)
1. `/0.58.6`: this will have the same output as `1`

> By "handled" I mean that they "came" for free, I didn't intend to support them but they work and I believe it's a good thing anyway.

## Test Plan

1. Run it locally;
1. Open the following links:
	- http://localhost:3000/?from=0.58.6&to=0.59.9
	- http://localhost:3000/0.58.6..0.59.9
	- http://localhost:3000/?from=0.58.6
	- http://localhost:3000/?to=0.59.9
	- http://localhost:3000/0.58.6


kelset:
> By "handled" I mean that they "came" for free, I didn't intend to support them but they work and I believe it's a good thing anyway.

💯

pvinis:
That's great! I have 2 things we could have before adding this:
- When I go to just localhost:8000, the url should not change. I think the url should change when I press "show me".
- Can we change `toVersion` to `to`, and same for `from`?

lucasbento:
@pvinis: those are good, gonna implement them!

lucasbento:
@pvinis: changes done, can you take a look? :)

src/components/common/VersionSelector.js
```diff
@@ -26,23 +27,63 @@ const ButtonContainer = styled.div`
   margin-top: 25px;
 `
 
+const getVersionsInURL = () => {
+  if (window.location.search) {
+    // Parses `/?from=VERSION&toVersion=VERSION` from URL
```
kelset:
```suggestion
    // Parses `/?from=VERSION&to=VERSION` from URL
```

lucasbento: Good catch, fixed!

PR #40, opened 2019-06-19
first description and comments for 60 release

pvinis:
# Summary
This PR adds some text for the upcoming 0.60 release.
The file should be renamed and the release name should too. Or at least duplicated?


src/releases/index.js
```diff
@@ -1 +1,6 @@
-export const versions = ['0.59.0', '0.58.0', '0.57.0']
+export const versions = [
+  '0.60.0-rc2',
```
lucasbento: I would put this as `0.60`, I believe this will be picked up as the logic uses `semver.coerce` to clean the version and remove pre-release suffix.

src/releases/0.60.0-rc2.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  comments: [
+    {
+      fileName: 'ios/Podfile',
+      lineNumber: 4,
+      lineChangeType: 'add',
+      content: (
+        <Fragment>
+          All these libraries below have been removed from the Xcode project file and now live in the Podfile.
+          Cocoapods handles the linking now.
```
lucasbento: Unfortunately new lines don't work in `markdown-to-jsx`: ([warning notice here](https://github.com/probablyup/markdown-to-jsx#usage) at the bottom).

src/releases/index.js
pvinis: I had to put it like this to make it show. I think it didn't otherwise. We would change is before merge.

src/releases/0.60.0-rc2.js
pvinis: I know, but it's still nicer in the code that one huge line, right? Is that what you're talking about? The comment is still just one line on the web.

src/releases/index.js
lucasbento: Interesting, do you think that we should fix that?

Btw the correct version would be `0.60.0-rc.2` as the `rc.1`: https://github.com/facebook/react-native/tree/v0.60.0-rc.1.

lucasbento: I mean, considering that we want people to upgrade to release candidates in order to provide feedback.

pvinis: That's true.. Hm, then maybe we could add these changes in this PR for rc1 and 2 as well?

src/releases/0.60.0-rc2.js
lucasbento: Oh yeah, makes sense!

src/releases/index.js
lucasbento: I think we should fix the check for the version so it shows for all the rc versions even if it's specified for a non-rc, what do you think?

pvinis: sounds good. lets talk on discord to make sure what the cases are and i can add it for this pr

lucasbento: Issue being tracked on #41.

lucasbento:
@pvinis: as #42 is done, what do we need to merge this?

PR #46, opened 2019-07-02
feat: move file after completed

### LINKED ISSUE #43: Move the file after it's completed
# Feature Request

This is related and started at #10, basically the website needs to show the "completed" files below the uncompleted ones.

## Why it is needed

It's easier this way to know which files are still left to be changed.

## Possible implementation

Have two sections of file _groups_, one for completed and another for uncompleted.

The project already stores the IDs of the completed ones [here](https://github.com/react-native-community/upgrade-helper/blob/643966b134372e962fe1e0518ccd50f474dd881b/src/components/common/DiffViewer.js#L26-L33).

KevinHu2014:
# Summary

This PR fixes #43 , it moves the completed files below the uncompleted ones. While mouseover on the green checkmark of a completed file,  ↩️  will be shown as a hint to reverse. After all the files are complete, the message "Your upgrade is done 🎉🎉" will appear.

## Test Plan

1. Run the PR locally
2. Open http://localhost:3000/?from=0.59.8&to=0.59.9
3. Complete files.

## Checklist

Demo2


lucasbento:
Just ran locally and this is pure ❤️ 

Gonna review the PR soon, thank you for doing this!

lucasbento:
@KevinHu2014: thank you so much for doing this! Just a couple things that I would suggest as changes:

1. I would remove the `Undone` title, it's the first thing the user sees after they click the `Show me how to upgrade` button and I believe there should be more focus in the diff, I'm totally onboard with the `Done` title though;
1. The `Your upgrade is done 🎉🎉` is a great addition but I believe it would be more suitable to be inside of an [`Alert`](https://ant.design/components/alert/) like this:

image

What do you think?

KevinHu2014:
@lucasbento  Thanks for the review 😊
Totally agree on both suggestions. 
I'll apply them and make the change as soon as possible.

KevinHu2014:
Ready for the second time review :)

image

src/components/common/DiffViewer.js
```diff
@@ -79,25 +83,59 @@ const DiffViewer = ({
...
-            diffKey={diffKey}
-            fromVersion={fromVersion}
-            toVersion={toVersion}
-            isDiffCompleted={completedDiffs.includes(diffKey)}
-            onCompleteDiff={handleCompleteDiff}
-            selectedChanges={selectedChanges}
-            onToggleChangeSelection={onToggleChangeSelection}
-          />
-        )
+        if (!completedDiffs.includes(diffKey)) {
```
lucasbento: I would suggest that instead of doing this condition and putting the component inside of you do:

```jsx
if (completedDiffs.includes(diffKey)) {
  return null;
}

return (
  <Diff
    key={`${diff.oldRevision}${diff.newRevision}`}
    {...diff}
    // otakustay/react-diff-view#49
    type={diff.type === 'new' ? 'add' : diff.type}
    diffKey={diffKey}
    fromVersion={fromVersion}
    toVersion={toVersion}
    isDiffCompleted={completedDiffs.includes(diffKey)}
    onCompleteDiff={handleCompleteDiff}
    selectedChanges={selectedChanges}
    onToggleChangeSelection={onToggleChangeSelection}
  />
)
```

So you don't need to have that much indentation on that many lines.

lucasbento: Also, on top of this, I would rather create a `DiffSection` component which receives a prop like `done={false}` and then renders the `Undone` section and the same goes the `Done`, so we don't have that much duplicated code.

src/components/common/DiffViewer.js
```diff
@@ -79,25 +83,59 @@ const DiffViewer = ({
...
+              onCompleteDiff={handleCompleteDiff}
+              selectedChanges={selectedChanges}
+              onToggleChangeSelection={onToggleChangeSelection}
+            />
+          )
+        }
+        return null
+      })}
+      {completedDiffs.length > 0 ? <Title>Done</Title> : null}
+      {diff.length === completedDiffs.length ? (
```
lucasbento: This statement and the one on the top would be more readable if separated into two components.

KevinHu2014:
@lucasbento Thanks for the review ❤️
Sorry I’m on a vacation, I will get back to this on Monday.

lucasbento:
@KevinHu2014 no rush! have a good vacation 🙂

src/components/common/DiffViewer.js
KevinHu2014: @lucasbento Do you mean something like this?

```js 
// Undone Section

<DiffSection done={false}>
  <Diff
    key={`${diff.oldRevision}${diff.newRevision}`}
    {...diff}
    // otakustay/react-diff-view#49
    type={diff.type === 'new' ? 'add' : diff.type}
    diffKey={diffKey}
    fromVersion={fromVersion}
    toVersion={toVersion}
    isDiffCompleted={completedDiffs.includes(diffKey)}
    onCompleteDiff={handleCompleteDiff}
    selectedChanges={selectedChanges}
    onToggleChangeSelection={onToggleChangeSelection}
  />
</DiffSection>


// Done Section

<DiffSection done={true}>
  <Diff
    key={`${diff.oldRevision}${diff.newRevision}`}
    {...diff}
    // otakustay/react-diff-view#49
    type={diff.type === 'new' ? 'add' : diff.type}
    diffKey={diffKey}
    fromVersion={fromVersion}
    toVersion={toVersion}
    isDiffCompleted={completedDiffs.includes(diffKey)}
    onCompleteDiff={handleCompleteDiff}
    selectedChanges={selectedChanges}
    onToggleChangeSelection={onToggleChangeSelection}
  />
</DiffSection>

```

Still a little confuse here, would need some guidance. 
Thanks:)

lucasbento: I mean that it would be nice to have the `DiffSection` component to decide which props should be passed to `Diff` based on the `done` prop.

For example, you have two `Diff` components now, what's the difference between them? You can get that difference (props) and put them inside of `DiffSection` which will only decide which prop to pass instead of having two `Diff` components with slightly the same code.

If you need any help around this please let me know and we will work on it together!

KevinHu2014: Thank you for the guidance.

The difference between them are whether are they in the `completedDiffs`.
Then I came up with the ideas below, trying to sort the Done one's after the Undone's.
```js
{diff.map(diff => {
        const diffKey = getDiffKey(diff)
        return (
          <DiffSection done={!completedDiffs.includes(diffKey)}>
            <Diff
              key={`${diff.oldRevision}${diff.newRevision}`}
              {...diff}
              // otakustay/react-diff-view#49
              type={diff.type === 'new' ? 'add' : diff.type}
              diffKey={diffKey}
              fromVersion={fromVersion}
              toVersion={toVersion}
              isDiffCompleted={completedDiffs.includes(diffKey)}
              onCompleteDiff={handleCompleteDiff}
              selectedChanges={selectedChanges}
              onToggleChangeSelection={onToggleChangeSelection}
            />
          </DiffSection>
        )
})}
```

But I'm stuck in this structure .
It seems to be like this.
```js
/*
 --DiffSection-------
|   Diff             |
--------------------
--DiffSection-------
|   Diff             |
--------------------
--DiffSection-------
|   Diff             |
--------------------
--DiffSection-------
|   Diff             |
--------------------
--DiffSection-------
|   Diff             |
--------------------
*/
```
Instead of this one

```js
/*
 --DiffSection-------
|   Diff             |
|   Diff             |
|   Diff             |
|   Diff             |
--------------------
--DiffSection-------
|   Diff             |
--------------------
*/
```

lucasbento: Alright, so, the way I envisioned this is as follows:

```jsx
const DiffSection = ({diff, getDiffKey, completedDiffs, isDoneSection}) =>
  diff.map(diffFile => {
    const diffKey = getDiffKey(diff);
    const isDiffCompleted = completedDiffs.includes(diffKey);

    // If the specific diff is done and this is being rendered in the
    // done section then we don't render anything
    if (!isDoneSection && isDiffCompleted) {
      return;
    }

    return (
      <Diff
        key={`${diff.oldRevision}${diff.newRevision}`}
        {...diff}
        // otakustay/react-diff-view#49
        type={diff.type === 'new' ? 'add' : diff.type}
        diffKey={diffKey}
        fromVersion={fromVersion}
        toVersion={toVersion}
        isDiffCompleted={completedDiffs.includes(diffKey)}
        onCompleteDiff={handleCompleteDiff}
        selectedChanges={selectedChanges}
        onToggleChangeSelection={onToggleChangeSelection}
      />
    );
  });
```

And then for the done section we just do:

```jsx
<DiffSection
  isDoneSection={true} // put this as false for the "undone" section
  diff={diff}
  getDiffKey={getDiffKey}
  completedDiffs={completedDiffs}
/>
```

lucasbento: Please let me know what you think about this.

We can also put the titles inside of the `DiffSection` so we get everything separated and barely code duplication.

KevinHu2014: Oh ! I see the concept now.  Thanks for your guidance 🙏

This occur to my mind before, but I was worrying about passing too many props down to the DiffSection. 

Just one more quick question, so we should also pass `fromVersion` , `toVersion` and `selectedChanges` as props right? Or there is some magical way that we could do this?

lucasbento: I would still pass them as props, yes.

Otherwise we would need to define it inside the main component and we don't really wanna go that path.

KevinHu2014:
@lucasbento Ready for the third time review :)

src/components/common/Diff/DiffSection.js
```diff
@@ -0,0 +1,61 @@
...
+    <div>
+      {title && completedDiffs.length > 0 ? <Title>{title}</Title> : null}
+      {diff.map(diffFile => {
+        const diffKey = getDiffKey(diffFile)
+        const isDiffCompleted = completedDiffs.includes(diffKey)
+
+        // If the specific diff is done and this is being rendered in the
+        // done section then we don't render anything
+
+        if (!isDoneSection) {
```
lucasbento: I don't really understand these conditions, what are they for?

src/components/common/Diff/DiffSection.js
```diff
@@ -0,0 +1,61 @@
...
+  title,
+  completedDiffs,
+  isDoneSection,
+  fromVersion,
+  toVersion,
+  handleCompleteDiff,
+  selectedChanges,
+  onToggleChangeSelection
+}) => {
+  return (
```
lucasbento: You can omit the `return` here as you can do `const func => () => (<div>...</div>)`.

src/components/common/DiffViewer.js
```diff
@@ -79,27 +80,39 @@ const DiffViewer = ({
...
+        diff={diff}
+        getDiffKey={getDiffKey}
+        completedDiffs={completedDiffs}
+        fromVersion={fromVersion}
+        toVersion={toVersion}
+        handleCompleteDiff={handleCompleteDiff}
+        selectedChanges={selectedChanges}
+        onToggleChangeSelection={onToggleChangeSelection}
+      />
+      {diff.length === completedDiffs.length ? (
```
lucasbento: I would suggest separating this in a different function.

Something like:

```jsx
renderUpgradeDoneMessage = ({ diff, completedDiffs }) =>
	diff.length === completedDiffs.length ? <Alert ...some props /> : null

// then inside of DiffViewer...

{renderUpgradeDoneMessage({ diff, completedDiffs })}
```

What do you think?

src/components/common/Diff/DiffSection.js
KevinHu2014: if ( ! isDoneSection && isDiffCompleted){}

If the isDoneSection is true. All the file will appear in the Done section. Because whatever the isDiffCompleted is, the condition will always be false.

For the conditions, outer ones is for distinguish whether its undone section or done section . The inner ones stands for what content should render.

src/components/common/Diff/DiffSection.js
KevinHu2014: Nice tip 👍🏻

src/components/common/DiffViewer.js
KevinHu2014: Nice suggestion 👍🏻

KevinHu2014:
I will add some space to it.👍🏻

src/components/common/Diff/DiffSection.js
lucasbento: I guess you could change to something like:

```
  // If it's the "done" section, it shouldn't show if it's not completed
  if (isDoneSection !== isDiffCompleted) {
    return
  }
```

What do you think?

KevinHu2014: wow that’s a nice one😎

KevinHu2014: wow that’s a nice one 😎

KevinHu2014:
@lucasbento Ready for final review :)

Issue #54, opened 2019-07-08
Configure ESLint and Prettier


lucasbento:
# Feature Request

Configure proper ESLint and Prettier, they both have their configs but should also be run on CircleCI together with the unit tests.

We already have the [lint script](https://github.com/react-native-community/upgrade-helper/blob/master/package.json#L31) but for some reason it's not picking up warnings/errors from Prettier.

lucasbento:
Enough work has been done in #60 so I'll close this.

Thank you @Jeanmsilva89!

PR #60, opened 2019-07-12
Fix eslint prettier config.

Jeanmsilva89:
# Summary

Addressing issue #54 by fixing the eslint config to use prettier. 

PS. I actually added the common eslint and prettier config from https://github.com/react-native-community/.github/blob/master/Guidelines/Repository%20Setup.md, however, it seemed like it was a bigger lift as there were quite a few places where the code would have to be updated. If the intention is to implement the common config on this repo, I'm happy to to update this PR to include it. 

## Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,7 @@
   "extends": "react-app",
   "plugins": ["prettier"],
   "rules": {
-    "jsx-a11y/accessible-emoji": 0
+    "jsx-a11y/accessible-emoji": 0,
+    "prettier/prettier": "error"
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,6 @@
   "extends": "react-app",
   "plugins": ["prettier"],
   "rules": {
-    "jsx-a11y/accessible-emoji": 0,
     "prettier/prettier": "error"
   }
 }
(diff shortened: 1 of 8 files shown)


.eslintrc
```diff
@@ -2,6 +2,7 @@
   "extends": "react-app",
   "plugins": ["prettier"],
   "rules": {
-    "jsx-a11y/accessible-emoji": 0
+    "jsx-a11y/accessible-emoji": 0,
```
lucasbento: I would rather not turn this off but fix the place that throws this warning, would you mind doing that?

src/__fixtures__/releases/0.59.js
```diff
@@ -1,11 +1,9 @@
 export default {
   usefulContent: {
-    description:
-      'This is the very nice 0.59 release!',
+    description: 'This is the very nice 0.59 release!',
     links: [
       {
-        title:
-          'This is a very cool link to a blog post about 0.59',
+        title: 'This is a very cool link to a blog post about 0.59',
```
lucasbento: I wonder why some trailing commas have been removed but this one did not, do you know the reason?

Jeanmsilva89: @lucasbento It's not a trailing comma, the url prop comes right after it.. :) took me a little while to see it.

lucasbento: That's true, so sorry I missed it.

kelset:
👏👏👏👏 thanks @Jeanmsilva89 !

PR #71, opened 2019-07-23
add a quick note about the wrong name change

pvinis:
Until we can just regenerate the diffs

# Summary

## Checklist

### DIFF
--- a/src/releases/0.60.js
+++ b/src/releases/0.60.js
@@ -39,7 +39,7 @@ export default {
       )
     },
     {
-      fileName: 'ios/RnDiffApp/Info.plist ',
+      fileName: 'ios/RnDiffApp/Info.plist',
       lineNumber: 8,
       lineChangeType: 'add',
       content: (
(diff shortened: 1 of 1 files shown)


lucasbento:
Isn't it better to manually delete the change on `rn-diff-purge`?

If not, I would rather add something to not render the diff for those files.

pvinis:
Ugh, it is, but at least for now we can have it there too, so people know what's up? I'll do a quick script to remove it from the diffs

src/releases/0.60.js
```diff
@@ -37,6 +37,26 @@ export default {
           for an explanation and some help with upgrading this file.
         </Fragment>
       )
-    }
+    },
+    {
+      fileName: 'ios/RnDiffApp/Info.plist ',
```
lucasbento: Can you remove the ending space on the file names?

pvinis: oopsssss

Issue #65, opened 2019-07-18
Add the ability to collapse/expand all files for easy navigation.


Jeanmsilva89:
# Feature Request

Allow users to collapse/expand all files at once. 

## Why it is needed

Makes it easier to find the right file on a large diff.

## Possible implementation

Hold `alt` and click on the inverted caret icon. It works as a toggle. 

https://github.blog/changelog/2018-08-17-collapse-all-diffs-in-a-pull-request-at-once/

kelset:
Ah yeah, I've been thinking about something like this - sounds like a good idea 🤗

If you can open a PR it would be great :)

lucasbento:
Whoa, that's pretty cool, would love to review a PR for this!

lucasbento:
Closing this as it has been done on #74, thanks again @Jeanmsilva89! 🙂

PR #74, opened 2019-07-26
Feat(toggleAll): add ability to toggle all files

Jeanmsilva89:
# Summary

Feature #65

Work in Progress

Add the ability to expand and collapse all files by holding the shift key and clicking on expand/collapse on any file. 
Since the Diff Section is split into 2 sections, this function works within the scope of the section you're interacting with. So if you use this function on the done section, only files in the done section will be affected and the same is true to files not in the done section.

## Checklist

### DIFF
--- a/src/components/common/Diff/DiffSection.js
+++ b/src/components/common/Diff/DiffSection.js
@@ -17,36 +17,54 @@ const DiffSection = ({
   handleCompleteDiff,
   selectedChanges,
   onToggleChangeSelection
-}) => (
-  <div>
-    {title && completedDiffs.length > 0 && <Title>{title}</Title>}
+}) => {
+  let diffRefMap = {}
 
-    {diff.map(diffFile => {
-      const diffKey = getDiffKey(diffFile)
-      const isDiffCompleted = completedDiffs.includes(diffKey)
-
-      // If it's the "done" section, it shouldn't show if it's not completed
-      if (isDoneSection !== isDiffCompleted) {
-        return null
+  const toggleAll = collapse => {
+    for (let diffKey in diffRefMap) {
+      let diffRef = diffRefMap[diffKey] && diffRefMap[diffKey].current
+      if (diffRef && diffRef.isDiffCollapsed !== collapse) {
+        diffRef.setIsDiffCollapsed(collapse)
       }
+    }
+  }
 
-      return (
-        <Diff
-          key={`${diffFile.oldRevision}${diffFile.newRevision}`}
-          {...diffFile}
-          // otakustay/react-diff-view#49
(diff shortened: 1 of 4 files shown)


src/components/common/Diff/DiffSection.js
```diff
@@ -17,36 +17,54 @@ const DiffSection = ({
...
+  let diffRefMap = {}
 
-    {diff.map(diffFile => {
-      const diffKey = getDiffKey(diffFile)
-      const isDiffCompleted = completedDiffs.includes(diffKey)
-
-      // If it's the "done" section, it shouldn't show if it's not completed
-      if (isDoneSection !== isDiffCompleted) {
-        return null
+  const toggleAll = collapse => {
```
lucasbento: What do you think about having one entry on `state` (something like `areAllDiffsCollapsed`) which, when set to `true`, collapsed all diffs automatically.

We can pass it to `Diff` and do `isDiffCollapsed={areAllDiffsCollapsed || isDiffCollapsed}`, that way we avoid calling all of them and have the state of this in one single place.

lucasbento:
This is really cool, thank you for doing this 🙌!

Just had a comment regarding the approach, please let me know what you think 🙂

src/components/common/Diff/DiffSection.js
Jeanmsilva89: @lucasbento It's my first time using hooks so I appreciate the feedback and also I'm taking my time to try different approaches. 

I initially didn't do what you suggested because it's a bit more complicated. Simply put if I add `isDiffCollapsed={areAllDiffsCollapsed || isDiffCollapsed}` it doesn't work because any time the `areAllDiffsCollapsed` is set **true**, it would take precedence over the local `isDiffCollapsed` and keep the Diff collapsed.

In order to get around this, I set up the `areAllDiffsCollapsed` I use `undefined`. This way I can set it to `undefined` any time I want to turn off the toggleAll functionality. For example, when a user clicks on a single Diff and breaks away from the `toggleAll` functionality I set it to undefined.

Anyways, it's a bit of workaround to get this work and I think it's after all cleaner than using ref. 

let me know what you think!

Jeanmsilva89:
@lucasbento I updated the PR to follow more-less what you suggested. And I also updated the trigger to `alt` key because that's what Github does. 

let me know your thoughts.

lucasbento:
@Jeanmsilva89: thank you, I'm going to review it between today and tomorrow!

lucasbento:
@Jeanmsilva89: is this ready to merge? just asking because of the `WIP` in the title.

Jeanmsilva89:
@lucasbento I updated the tile. You can merge it.

lucasbento:
Thank you so much for the contribution, @Jeanmsilva89!

Issue #75, opened 2019-07-26
Add eslint & prettier pre git hooks


vinnyA3:
# Feature Request

I think it would be helpful to have a precommit / post hooks to run eslint and/or prettier

## Why it is needed

This way, code that doesn't conform to repo style standards won't push.  Also helps with automating linting & prettifying tasks.

## Possible implementation

Use [`husky`](https://www.npmjs.com/package/husky) package to create pre hooks

vinnyA3:
This issue can be closed 🎊

PR #76, opened 2019-07-28
feat: add prehooks

vinnyA3:
# Summary

This PR--for issue (#75)--helps to automate linting & prettifying tasks--which ultimately helps to maintain stylistic repo conventions.

I implemented this feature by:
  * adding the `(husky)[https://www.npmjs.com/package/husky]` and `(pretty-quick)[https://www.npmjs.com/package/pretty-quick]` as dev dependencies
  * added appropriate `package.json` configuration

## Test Plan

* make changes to a `js` file
* `git commit <file>`
* `git push`

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,12 @@
     "test": "react-scripts test",
     "lint": "eslint . --cache --report-unused-disable-directives"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged --pattern \"src/**/*.*(js|jsx)\"",
+      "pre-push": "npm run lint"
+    }
+  },
   "browserslist": {
     "production": [
       ">0.2%",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "pretty-quick --staged --pattern \"src/**/*.*(js|jsx)\"",
-      "pre-push": "npm run lint"
+      "pre-push": "yarn run lint"
     }
   },
   "browserslist": {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,9 @@
     "eslint": "^5.16.0",
     "eslint-plugin-prettier": "^3.1.0",
     "gh-pages": "^2.0.0",
-    "prettier": "^1.18.2"
+    "husky": "^3.0.1",
+    "prettier": "^1.18.2",
+    "pretty-quick": "^1.11.1"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1327,6 +1327,11 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-12.0.2.tgz#3452a24edf9fea138b48fad4a0a028a683da1e40"
   integrity sha512-5tabW/i+9mhrfEOUcLDu2xBPsHJ+X5Orqy9FKpale3SjDA17j5AEpYq5vfy3oAeAHGcvANRCO3NV3d2D6q3NiA==
 
+"@types/normalize-package-data@^2.4.0":
+  version "2.4.0"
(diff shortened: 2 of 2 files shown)


lucasbento:
This looks very nice, do you know how much time was increased when doing the commit with the hook?

I'm usually hesitant to do it as it increases time considerably (not sure if only in big projects) for commit and push, I would rather have this on the CircleCI only and explicitly mentioned on a contributing guide.

Thoughts on this @vinnyA3 & @pvinis?

pvinis:
I don't mind having it. I want one thing thought, to **not** run in the `diffs` branch or the `app-base` or any of the `release/*` branches. Do you know if that's possible with husky? I wouldn't want it to run in the generated files, in case we have a `src` dir at some point.

vinnyA3:
I don't know exactly how much time increased when using the hook; but for a couple of files, the increase seems negligible.  I work on a large project using an eslint precommit hook, and haven't noticed any debilitating delay; however, since `prettier` is running precommit, I would imagine things would be slower since entire files are formatted.  That said, prettier is relatively efficient.  If you'd like, I can definitely run some performance tests.

@pvinis One way to run hooks on specific branches would be to create a small bash script (the script would check which branch is currently checked out).  Another way would be to include the `git-branch-is` package, or something similar, to run the check inside the hook command itself.  So, something like:
```
"hooks": {
    "pre-commit": "if git-branch-is master .... "
}
```

The aforementioned should work; I will test this.

lucasbento:
> I don't know exactly how much time increased when using the hook; but for a couple of files, the increase seems negligible. I work on a large project using an eslint precommit hook, and haven't noticed any debilitating delay; however, since prettier is running precommit, I would imagine things would be slower since entire files are formatted. That said, prettier is relatively efficient. If you'd like, I can definitely run some performance tests.

Okay, I'm sold.

@pvinis: can you take a look at @vinnyA3's suggestion regarding the branch?

pvinis:
I will look into it. Both for the branch and the performance actually.

(I like adding some extra dis comments on my comments, so here is one. I don't like prettier 🌚 )

package.json
```diff
@@ -22,14 +22,22 @@
...
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "lint": "eslint . --cache --report-unused-disable-directives"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged --pattern \"src/**/*.*(js|jsx)\"",
```
pvinis: is this `*.*` needed? can't we just have `.*` if its a regex, or `*.` if its a glob?

pvinis:
I actually tried `git-branch-is` and it worked great, then I realized we are in the upgrade-helper and not the purge, so my concern was invalid :D 
In any case, now we know it works, if we want it in the future.

For performance, eh it's fine. Any hooks adds a tiny bit of delay, but I don't think it's so bad or intrusive.

I only have two things, one is to change to yarn, and the other is about the regex/glob.

package.json
```diff
@@ -22,14 +22,22 @@
...
+      "pre-push": "npm run lint"
```
pvinis:
```suggestion
       "pre-push": "yarn lint" 
```
Since we use yarn for everything else.

package.json
vinnyA3: Yes, that should work fine!

vinnyA3:
@pvinis will make the changes today

package.json
vinnyA3: @pvinis So I tested your suggestions, and couldn't get that to work.  `pretty-quick` uses the `minimatch` package internally--which seems to use a traditional glob--but was failing to expand without `*.*`.  Another approach would be to use two `--pattern` options, but that seems a bit verbose.  Let me know what you think!

vinnyA3:
Sounds good, if I find a way around it, I can create another PR

PR #80, opened 2019-07-31
Add "view file" link

### LINKED ISSUE #79: View full file contents
# Feature request

Instead of viewing just the diff between different file version, why not just allow the user to view the full contents of the file? I.e. "view raw"

## Why it is needed

Some diffs are just not that useful 🤷‍♀ 

For instance, the `.flowconfig`-file is (in my case) never manually edited. The diff is not very helpful in this case, because all I want is just the "new version of the file".

## Possible implementation

Add a "view raw" or "view file" to the "header" of the diff.

image

mausworks:
# Summary

This PR adds a "View file" link to each file diff header when there is a diff available.

This feature is useful for some configuration files which do not need be manually merged, but could rather just be overwritten.

Another contending solution would be to always display the "Download"-button.

But, as of now, it works like this:

- If the file has a diff, the "View file" link is shown
- If the file does not have a diff the "Download" button is shown instead (i.e. for non text-files)

I think this is useful, because it shows clearer intent.

Closes #79 

## Test Plan

Screenshot:
image

I copied the `DownloadFileButton` component, and changed it accordingly.

## Checklist

### DIFF
--- a/src/components/common/Diff/DiffHeader.js
+++ b/src/components/common/Diff/DiffHeader.js
@@ -86,6 +86,25 @@ const DownloadFileButton = styled(({ visible, toVersion, newPath, ...props }) =>
   }
 `
 
+const ViewFileButton = styled(({ visible, toVersion, newPath, ...props }) =>
+  visible ? (
+    <Button
+      {...props}
+      type="link"
+      target="_blank"
+      href={getBinaryFileURL({
+        version: toVersion,
+        path: newPath
+      })}
+    >
+      View file
+    </Button>
+  ) : null
+)`
+  font-size: 12px;
+  color: #24292e;
+`
+
 const CompleteDiffButton = styled(
   ({ diffKey, isDiffCompleted, onCompleteDiff, ...props }) =>
     isDiffCompleted ? (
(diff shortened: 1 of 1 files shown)


mausworks:
@pvinis Please have a look at my latest commit as well. :)

pvinis:
Checking now. (Oh god, our code is getting uglier and uglier with `styled` 😆)

pvinis:
Looks good. I'll leave it for lucas to also check, and we can merge after.

Thank you so much for this PR :D

mausworks:
> Looks good. I'll leave it for lucas to also check, and we can merge after.

Great! Can't wait for it :)

> Thank you so much for this PR :D

You are welcome.

mausworks:
> Checking now. (Oh god, our code is getting uglier and uglier with styled 😆)

Haha, yeah, the components are getting pretty long! I can take it upon myself to do some refactoring, but at a later time.

I might in that case change some of the project structure. Is that OK?

pvinis:
Feel free to play around. It would be great to also make an issue or a PR when you have an idea (if it's a big change), so we can contribute early. Not that you change everything around to something you find awesome and then we start complaining. :p

So yes, feel free to play around, suggest ideas. I will not have much time in the next couple of months to refactor things, otherwise I would also look into it. But I will still be here! :)

mausworks:
Actually, hold this PR for a sec. I forgot to handle the case where the new file is deleted.

(I'm also pretty sure that the "Download file" has this bug as well)

Fixing ASAP.

Issue #17, opened 2019-06-08
Add unified diff option


lucasbento:
# Feature Request

Some people may prefer to see the diff unified like GitHub provides, we should offer an option to do so.

kelset:
True, but I feel that it's not a blocker for v1

lucasbento:
No worries, we will stick to #10.

iamrestrepo:
I can take this if nobody is working on it already.  😃

iamrestrepo:
Done ✅  let me know if it requires any change/update. 😃

lucasbento:
Fixed by @iamrestrepo on #87 🎉

PR #87, opened 2019-08-12
feat: adds style viewing options for differences

iamrestrepo:
# Summary
 This PR-- aims to solve issue (#17 )--Add unified diff option

I implemented this feature by:

- Implementing a DiffViewStyleOptions component to render the viewing options 
- Applying viewing options to diff component

Screen Shot 2019-08-13 at 10 37 21 AM

Screen Shot 2019-08-13 at 10 37 37 AM

## Checklist

### DIFF
--- a/src/components/common/Diff/DiffViewStyleOptions.js
+++ b/src/components/common/Diff/DiffViewStyleOptions.js
@@ -1,13 +1,11 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Button } from 'antd'
-
-const Options = Button.Group
+import { Radio } from 'antd'
 
 const Container = styled.div`
   position: relative;
   border-width: 1px;
-  margin-top: 10px;
+  margin-top: 16px;
   flex-direction: row-reverse;
   display: flex;
 `
@@ -15,10 +13,14 @@ const Container = styled.div`
 const DiffViewStyleOptions = styled(({ setViewStyle, ...props }) => {
   return (
     <Container>
-      <Options>
-        <Button onClick={() => setViewStyle('split')}>Split</Button>
-        <Button onClick={() => setViewStyle('unified')}>Unified</Button>
-      </Options>
+      <Radio.Group>
+        <Radio.Button value="split" onChange={() => setViewStyle('split')}>
+          Split
+        </Radio.Button>
+        <Radio.Button value="unified" onChange={() => setViewStyle('unified')}>
+          Unified
+        </Radio.Button>
+      </Radio.Group>
     </Container>
   )
 })`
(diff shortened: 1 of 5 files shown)


kelset:
This looks awesome, thanks @iamrestrepo ❤️

Can you please add a screenshot or a GIF? 😇

src/components/common/Diff/DiffViewStyleOptions.js
```diff
@@ -0,0 +1,40 @@
...
+  border-width: 1px;
+  margin-top: 10px;
+  flex-direction: row-reverse;
+  display: flex;
+`
+
+const DiffViewStyleOptions = styled(({ setViewStyle, ...props }) => {
+  return (
+    <Container>
+      <Options>
```
lucasbento: I believe we can have this as `Button.Group` as there's no change on the style.

lucasbento: Instead of assigning Button.Group to Options (as it's done on line 5), you can just use the Button.Group here.

iamrestrepo: Got you @lucasbento.  Changes have been done!

lucasbento:
@iamrestrepo: the default behaviour for the view style seems to be unified now, what do you think about putting it as split (as it's the current behaviour) and save to [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) whatever the user selects?

Also, it would be nice to have the button on an active state for the initial style.

Please let me know your thoughts!

iamrestrepo:
Sounds good @lucasbento. Will get to it by the end of the week!

iamrestrepo:
All done ✅!

src/components/common/Diff/DiffViewStyleOptions.js
```diff
@@ -0,0 +1,52 @@
...
+const Container = styled.div`
+  position: relative;
+  border-width: 1px;
+  margin-top: 16px;
+  flex-direction: row-reverse;
+  display: flex;
+`
+
+const DiffViewStyleOptions = styled(({ setViewStyle, ...props }) => {
+  const [viewStyle, changeActiveView] = useState(
```
lucasbento: Is there any particular reason that you need to define this here even though it's also defined on `DiffViewer`?

I would suggest moving this there together with the `localStorage` handling and then this is good to go.

lucasbento:
@iamrestrepo: there's an error on the build preview, you can [see it here](https://app.netlify.com/sites/upgrade-helper/deploys/5d5c678af74c4700083a75ce), the problem apparently is:

image

iamrestrepo:
> @iamrestrepo: there's an error on the build preview, you can [see it here](https://app.netlify.com/sites/upgrade-helper/deploys/5d5c678af74c4700083a75ce), the problem apparently is:
> 
> image
> 
> Can you fix it?

Done! ✅

src/components/common/Diff/DiffViewStyleOptions.js
iamrestrepo: ✅

PR #89, opened 2019-08-13
Only run `deploy` on branches on the original repository

### LINKED ISSUE #47: React Native 0.60.0 `xcodeproj` upgrade
# Moved here: https://github.com/react-native-community/upgrade-support/issues/14

--

For an easy upgrade experience of this file, do the following.

After all the other files have been upgraded, run `yarn install` and
```
cd ios
pod install
cd ..
```

This will cause all the Pods to be connected. After that, open the workspace file in the `ios` directory, called something like `MyApp.xcworkspace`

Screenshot 2019-07-03 at 16 07 01

This workspace contains your original project, as well as the `Pods` project. Expand your project and find the `Libraries` directory.

Screenshot 2019-07-03 at 16 07 16

Inside that, there are a few libraries related to React Native, and possibly other native modules from other libraries you might have connected.

Screenshot 2019-07-03 at 16 07 27

You can go ahead and remove the React Native related libraries from there, by selecting them and hitting the delete button. In the popup you get, you can select `Remove references`. If you select to `Move it to Trash`, you might need to run `yarn install` again.

Screenshot 2019-07-03 at 16 07 40

These libraries are not needed there anymore, since now they are connected using CocoaPods. You can do the same for other libraries that might exist in that directory, if they support CocoaPods. If they don't, keep them there for now.

After all this, your diff of that file should look roughly the same as the one right here. Don't worry too much if it doesn't match exactly, as your project might have more or fewer targets.

Try to see if the rough outline is the same, and of course make sure your project can build and run.

lucasbento:
This PR fixes #47, it disallow the `deploy` job to be run on forks, after this is merged we can enable the setting on CircleCI to run jobs on forks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,8 @@ workflows:
             branches:
               only:
                 - master
+              # Forked pull requests have `CIRCLE_BRANCH` set to pull/XXX
+              ignore: /pull\/[0-9]+/
 
       - deploy:
           requires:


lucasbento:
I'm merging this as we sorta need this to work asap, we will iterate over it later if needed.

Issue #45, opened 2019-07-02
Add a "how to use" section


lucasbento:
# Feature Request

Working on facebook/react-native-website#1063 I noticed that we lack a "getting started"/"how to use" guide for the user to know what to do once they land on the page.

## Why it is needed

So it's more clear to the user what they need to do with that beautiful diff that's showing in their browser.

## Possible implementation

A button with a `?` icon that opens a modal with some nice content about how to upgrade using Upgrade Helper, content can be based on facebook/react-native-website#1063.

kelset:
Yeah, this sounds like a good approach - probably just a few lines explaining👍

KevinHu2014:
Could I have this one :)


By the way, Where should this button belongs?
Will it be nice if placing it besides the Github star badge button?
> A button with a ? icon that opens a modal with some nice content about how to upgrade using Upgrade Helper

lucasbento:
@KevinHu2014: of course!

I was thinking more like putting it on the top right of the container, something like here:

<img width="654" alt="Screenshot 2019-07-03 at 09 52 50" src="https://user-images.githubusercontent.com/6207220/60573663-5d644700-9d78-11e9-92f9-73a1eabccf83.png">

But I suppose it would also look nice close to the star button, any of the two solutions would work well in my opinion.

KevinHu2014:
Sorry I forgot this one. 😓

@lucasbento Shall we discuss the short version content of "How to use" first?
And since https://github.com/react-native-community/upgrade-helper/pull/91#issuecomment-524590639, the how to button still goes to top right of the container right?

vinnyA3:
Hi @KevinHu2014

I think the top right corner is a nice position.  It can go right next to the settings button. What do you think @lucasbento?

lucasbento:
> I think the top right corner is a nice position. It can go right next to the settings button. What do you think @lucasbento?

Oh yeah, definitely agree with this!

I would start doing this after #91 is merged, @KevinHu2014.

pvinis:
The talk is out where I explain how to use this. Should we link to that? 😬

kelset:
100% :)

lucasbento:
> The talk is out where I explain how to use this. Should we link to that? 😬

Would be nice!

@KevinHu2014: you can also use https://facebook.github.io/react-native/docs/upgrading#upgrade-helper as reference.

PR #91, opened 2019-08-16
feat: show latest `rc` versions in `from` dropdown

### LINKED ISSUE #49: Show `rc` versions in the `from` dropdown
We show `rc` versions on the `to` dropdown until the final release is out.
After the final is out, if a user is on an `rc` version, they should be able to upgrade from that to the final.

Therefore we should show the `rc` versions on the `from` dropdown.

(If we want to minimize noise, we should only show the `rc`s of the latest version.)

vinnyA3:
# Summary
The *from* dropdown menu now contains *rc* versions of the latest release 🎊!
Prior to this PR, `upgrade-helper` filtered out *rc* versions on the *from* dropdown; however, if a user was on an *rc* version looking to migrate, they wouldn't have the steps to upgrade to the latest release.

To reduce noise, only *rc* versions of the latest release are displayed.

Resolves #49 

## Test Plan

Since the latest release does not have any prior release candidates, there aren't any to display in the dropdown.  To test the functionality, I added mock *rc* versions to the list.

**mock insertion code**:
```
  // VersionSelector.js
  // inside function 'isFromAValidReleaseCandidate' ....

  releasedVersions.unshift('0.60.5-rc')

  for (let i = 1; i < 4; i++) {
    releasedVersions.unshift(`0.60.5-rc${i}`)
  }
```

## Screenshots

**'from' dropdown with latest version's rcs**:
Screen Shot 2019-08-16 at 8 06 36 PM

---

**'to' dropdown does not contain any release candidates**:
Screen Shot 2019-08-17 at 2 22 57 PM

## Checklist


lucasbento:
If the release has lots of release candidates, it may create a lot of noise to the users that don't need to see it.

We could have a configuration button somewhere on the top of the first `div` that opens a popover with _soon to be_ multiple config buttons, we could have a toggle for `rc` versions there.

What do you think? Apologies for throwing the idea just now, it's a suggestion that maybe could improve the UX.

pvinis:
Like a "show latest release candidates" checkbox? That could be nice, yea!

vinnyA3:
> `0.60` is the major release and if it has `0.60.0-rc.1`, it should be shown in the dropdown

@lucasbento Ahh! yes, I see.  That makes sense.  That should be a matter of adding an additional check to the helper function I made ie. `if fromVersion is a majorRelease && ....`.  Will make that adjustment

src/components/common/VersionSelector.js
```diff
@@ -160,11 +172,14 @@ const VersionSelector = ({ showDiff }) => {
...
-        toVersion: toVersionToBeSet,
-        latestVersion
-      })
+      const sanitizedVersions = getReleasedVersionsWithCandidates(
+        {
+          releasedVersions: allVersionsFromResponse,
+          toVersion: toVersionToBeSet,
+          latestVersion
+        },
+        true
```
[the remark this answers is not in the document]
vinnyA3: Thanks for the catch.  Originally, I had a check to see if the `isFromVersionAReleaseCandidate` should be applied

src/components/common/VersionSelector.js
```diff
@@ -54,19 +54,28 @@ const compareReleaseCandidateVersions = ({ version, versionToCompare }) =>
...
   latestVersion
 }) => {
   const isToVersionAReleaseCandidate = semver.prerelease(toVersion) !== null
   const isLatestAReleaseCandidate = semver.prerelease(latestVersion) !== null
 
-  return releasedVersions.filter(
-    releasedVersion =>
+  return releasedVersions.filter(releasedVersion => {
+    return (
+      isFromAValidReleaseCandidate({
```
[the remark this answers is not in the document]
vinnyA3: 😄 glad it's working nicely

vinnyA3:
@lucasbento @pvinis 

The popover UI component sounds like a nice feature (especially if there are more features to be added to the popover).  My first thought is that it would be located in the top right of the main component.  Is this the `div` that you were referring to?

The checkbox toggle inside the popover would be great.  Do you think this should be part of the same issue? Or split out separately?

lucasbento:
@vinnyA3: that looks pretty damn good!

I was thinking of some here:

Screenshot 2019-08-21 at 09 30 34

> The checkbox toggle inside the popover would be great. Do you think this should be part of the same issue? Or split out separately?

I do realize that it's not ideal to increase the scope of the issue while you are doing it but I think the ending result would be better.

This is also a related issue #45.

vinnyA3:
That sounds good, I can implement later today, or tomorrow.
I guess my only question(s) are: when the user toggles the checkbox (show release candidates) inside of the popover, only the 'latest' rcs are displayed (the functionality this PR addresses); and, is the 'how to use' section still a feature?  Just curious, as having two buttons there (a question mark and settings 'gear'), in my opinion, would get a little cramped.

lucasbento:
@vinnyA3: that's a good question.

My idea is to have a button (like a cog or something) and that opens a popover with multiple buttons inside of it and with a text on the right side of each one them so it's not confusing.

vinnyA3:
@lucasbento @pvinis  Created the settings dropdown functionality.  The latest release candidates are shown in the dropdown by default. Let me know what you think!
settings-fn

src/utils.js
```diff
@@ -33,3 +33,6 @@ export const getChangelogURL = ({ version }) =>
     '.',
     ''
   )}`
+
+// settings constants
+export const SHOW_LATEST_RCS = 'show latest release candidates'
```
lucasbento: Why does it need to be separated here?

src/components/pages/Home.js
```diff
@@ -43,12 +45,16 @@ const StarButton = styled(({ className, ...props }) => (
...
   margin-left: 15px;
+  margin-right: auto;
 `
 
 const Home = () => {
   const [fromVersion, setFromVersion] = useState('')
   const [toVersion, setToVersion] = useState('')
   const [showDiff, setShowDiff] = useState(false)
+  const [settings, setSettings] = useState({
+    [`${SHOW_LATEST_RCS}`]: true
```
lucasbento: I'm not sure if the default here should be `true` or not as we are currently not showing it and not all the user will need to see the `rc` versions.

Thoughts on this @vinnyA3 & @pvinis?

pvinis:
I tried changing the code to
```
  const allVersionsFromResponse = [
        '0.61.0-rc.2',
        '0.61.0-rc.1',
        '0.60.4',
        '0.60.4-rc.1',
        '0.60.2',
        '0.60.0',
        '0.59.3'
      ]
```
to see how it would show up with just a new patch rc, then a major rc, and i got some funky results :p 

Probably a patch rc we can skip, but for example with the 2 rcs on 61, i cannot choose to upgrade from 60.4 to 61 rc1, I can only see rc2.

lucasbento:
> Probably a patch rc we can skip, but for example with the 2 rcs on 61, i cannot choose to upgrade from 60.4 to 61 rc1, I can only see rc2.

@pvinis: there may be a bug there but just to be clear: **there are currently no release candidates for patches, only for majors**.

pvinis:
Right, so we can forget the patch rc line i had in the code there, but we should still prepare it for a major/minor/whatever the middle number is :p for 61 basically.

src/components/pages/Home.js
vinnyA3: I agree.  Having the ability to toggle that when needed is good

lucasbento:
@vinnyA3: let us know when it's ready for review 😁

src/utils.js
vinnyA3: @lucasbento I figured since the string 'show latest release candidates' is used in multiple components (Home, Settings), it would be nicer to have a constant.  It's used twice in Home and a few times in Settings.  Instead, I could hard code the string in Home and just keep the constant local to `Settings.js`.  Let me know what you think!

vinnyA3:
@lucasbento  @pvinis  Ready! sorry for the delay :(

pvinis:
Ok first, no need to say sorry :D 

Second, I can only see these in `from` 
Screenshot 2019-09-16 at 10 03 42
Did something get messed up in the version filtering?

pvinis:
I might need to spend some time adding a view tests for the version selection functions, so we have this nice and stable for everyone :)

lucasbento:
@pvinis: I think it's because the latest is currently a `rc`.

@vinnyA3: what's the behaviour that we should be testing here? with the latest being a release candidate, should there be another `rc` on `from`?

vinnyA3:
> @vinnyA3: what's the behaviour that we should be testing here? with the latest being a release candidate, should there be another `rc` on `from`?

@lucasbento Well, originally I was working on the logic to display the RC's of the latest version.  I didn't think of displaying if there were new RCs available; however, I think it would be nice to include both the RCs of the latest major version && the new available RCs.  What do you think about this?

lucasbento:
@vinnyA3: sorry for the delay on responding this.

> I think it would be nice to include both the RCs of the latest major version && the new available RCs. What do you think about this?

I can't say that I have a definitive response to this but I think we should try it.

I'm a little bit afraid of clogging the UI with a bunch of versions but I believe we should help anyway the people that are on `rc`s.

lucasbento:
@vinnyA3: I did some minor changes to get it to work with the latest release and I believe this is good to go, let's keep talking if we want to iterate on top of this.

Thank you so much for working on this!

PR #102, opened 2019-10-12
Add Decoration to DiffView

### LINKED ISSUE #101: Diff doesn't show section changes / misleads on where changes appear
# Bug
As screen in this screenshot, the context for the `!debug.keystone` change shows under the `"# IntelliJ"` section which is quite misleading. By looking at the line numbers you may notice that they are not linear - **there is a jump from line 25 to 39**. However there is no notice of this, so the context lines appears together with the wrong section header.

Screen Shot 2019-10-04 at 13 51 53

Since some changes are context sensitive, it may cause problems for people while upgrading. In this particular example (`.gitignore`) the context isn't important beyond readability and future upgrade diffs.

## React Native versions
0.60.6 => 0.61.2

Screen Shot 2019-10-04 at 13 56 47

## Steps to reproduce
1. Open https://react-native-community.github.io/upgrade-helper/?from=0.60.6&to=0.61.2
2. Scroll down to the diff of .gitignore
3. See that "Android/IntelliJ" appears as the header for the Buck ignore changes.

Describe what you expected to happen:
There should be a visual divider or other visual clue when line numbers increment more than 1.

KevinHu2014:
# Summary

This fixes #101, 
Without the decoration in hunk might lead to some misunderstanding.
In this PR decoration is added to make it more clear to the user.

## Test Plan

1. Run the PR locally
2. Open http://localhost:3000/?from=0.60.6&to=0.61.2
3. Check .gitignore

## Checklist

螢幕快照 2019-10-12 上午10 58 32

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -1,6 +1,12 @@
 import React, { useState } from 'react'
 import styled from 'styled-components'
-import { Diff as RDiff, Hunk, markEdits, tokenize } from 'react-diff-view'
+import {
+  Diff as RDiff,
+  Hunk,
+  markEdits,
+  tokenize,
+  Decoration as DiffDecoration
+} from 'react-diff-view'
 import DiffHeader from './DiffHeader'
 import { getComments } from './DiffComment'
 
@@ -9,6 +15,17 @@ const Container = styled.div`
   border-radius: 3px;
   margin-bottom: 16px;
   margin-top: 16px;
+`
+
+const More = styled.div`
+  background-color: #f1f8ff;
+  margin-left: 30px;
+  padding-left: 4px;
+  color: '#1b1f23b3';
+`
+
+const Decoration = styled(DiffDecoration)`
+  background-color: #dbedff;
 `
 
 const DiffView = styled(RDiff)`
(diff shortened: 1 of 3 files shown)


pvinis:
Awesome job! Thank you so much.

src/components/common/Diff/Diff.js
```diff
@@ -122,14 +139,17 @@ const Diff = ({
...
+                <More>{hunk.content}</More>
+              </Decoration>,
               <Hunk
                 key={hunk.content}
                 hunk={hunk}
                 tokens={tokens}
                 gutterEvents={{ onClick: onToggleChangeSelection }}
               />
-            ))
+            ])
```
pvinis: Can I suggest we change from `[ <Dec />, <Hunk /> ]` to `<> <Dec /> <Hunk /> </>`? It seems cleaner.

KevinHu2014: At first I used `<> <Dec /> <Hunk /> </>`, but get an error of `provide key id to elements is need`. 
So I change to `[ <Dec />, <Hunk /> ]` 😅
 Is there any easy workaround for adding key to`<></>` ?

pvinis: 🤦‍♂️

Ok, let's keep it as is then.

PR #107, opened 2019-12-18
some comments on 62

pvinis:
# Summary

## Checklist

### DIFF
--- a/src/releases/index.js
+++ b/src/releases/index.js
@@ -1,4 +1,5 @@
-const versions = ['0.61', '0.60', '0.59', '0.58', '0.57']
+// this line HAAAAAAAAS to go
+const versions = ['0.62', '0.61', '0.60', '0.59', '0.58', '0.57']
 
 export default versions.map(version => ({
   ...require(`./${version}`).default,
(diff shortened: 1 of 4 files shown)


src/releases/index.js
```diff
@@ -1,4 +1,5 @@
-const versions = ['0.61', '0.60', '0.59', '0.58', '0.57']
+// this line HAAAAAAAAS to go
```
lucasbento: IMO this comment has to be an issue, not a comment.

Why does this line has to go? what does it need to be replaced with?

pvinis: indeed, I just quickly made it there. I'll add it to the rework issue.

I was trying to add some comment and links etc for 62 and I forgot to add it here, and they would not show up. I want to make this part go away. if there is a comment or link or whatever for a release, it should be shown without the need for manual addition in this file and list, you know?

pvinis: https://github.com/react-native-community/upgrade-helper/issues/108
the second point is touching this part, so this comment is now not needed I guess, but we can leave it as a reminder to rework it soon 😁 I started a branch but on holidays it's been slow.

Issue #108, opened 2019-12-18
Cleanup/Refactor Goal 


pvinis:
# Description

- [ ] We need to refactor the code here, it has become too complex for such a simple project 😬, but I haven't had the time to do it myself yet. Basically I want to have a simple utility thing that fetches the releases, and then generates a list of the versions to be displayed. Then we will have the ui thing that will actually display all that. Then we can actually test the utility part, and the ui I don't care if we test.

- [ ] Also, we need to make the notes/comments/links better. Currently, we can only comment inline on minor versions (as it 0.59.x), but if a file has been touched on 0.59.1 and reverted on 0.59.2, this comment is now broken. We need to make a better plan for this.

- [ ] We got a great idea on twitter https://twitter.com/safaiyeh/status/1205969341969133568?s=20 about adding a text input field that the user can put their app name, and all instances of `RnDiffApp` will be replaced with that name. Sounds cool, helpful and easy, so let's do it.

- [ ] Write a document about xcodeproj. I explained it in my talk, but I should also write a text and have it here so its easy to find. For now, general explanations are needed. No specific questions to make an FAQ or something.

safaiyeh:
Nice!! @Naturalclar thanks for that feature.

kelset:
Hey folks — I’m doing a maintenance pass and catching up on some of the older backlog.

I think this one has outlived its usefulness as an open tracker. It ended up bundling a bunch of different ideas together, and parts of it are already done now anyway — for example the app-name replacement work landed a long time ago, and the TypeScript migration happened too.

What’s left here is either too broad or better tracked as separate focused issues when we actually want to work on them.

So I’m going to close this one out. If we still want to tackle any of the remaining items, let’s open dedicated issues for those specific pieces of work so they’re easier to scope and review.

PR #111, opened 2019-12-27
Feature: replace RnDiffApp with entered AppName

Naturalclar:
# Summary

This PR is part of #108 
This adds a text input in the header which allows user to enter their AppName.
All the `RnDiffApp` and `rndiffapp` will be replaced with the entered AppName (in lowercase if needed)

## Test Plan

Kapture 2019-12-27 at 20 39 44

## Checklist


pvinis:
so good! awesome. I'll test it when I'm home before merging.

lucasbento:
Hi @Naturalclar, thank you for opening this PR.

The functionality is pretty useful, however, I would rather have the input for the application name to be under a button (the cog on the top right, for instance) as this is not required for everyone and can be just noise for most of the users.

About the warning on the bottom, I like it, I would suggest having it:

- Only showing when the user clicks on the `Show me how to upgrade!` button;
- As an info alert, [you can check it here](https://ant.design/components/alert/);
- With a button to close it which will only show it again if the user reloads/reopens the page (using state).

What do you think?

Naturalclar:
@lucasbento @pvinis 
Thanks for the feedback!
I've placed the AppName input inside the setting PopOver.
What do you think?

Kapture 2019-12-29 at 22 55 01

As for the warning on the bottom, it's not part of this PR, but I'll gladly make another PR that makes the change you request for :)

pvinis:
Pretty great! I have two things to comment on.

- I would like a placeholder for the input field, some like `MyAwesomeApp`, so users can see the usual capitalization.
- When I type fast in there, the whole diff view refreshes and some of the time the result is not correct, like
Screen Shot 2019-12-30 at 10 48 10

Naturalclar:
> I would like a placeholder for the input field, some like MyAwesomeApp, so users can see the usual capitalization.

Sounds good! I've added a placeholder to the input form.

Screen Shot 2019-12-31 at 2 23 42 PM

>When I type fast in there, the whole diff view refreshes and some of the time the result is not correct

I'm wondering what would be the best way to make the form without this error happening.
Should I place a buton inside settings to set the AppName so that refetching won't happen during input?

src/components/common/DiffViewer.js
```diff
@@ -61,6 +62,16 @@ const DiffViewer = ({
     localStorage.setItem('viewStyle', newViewStyle)
   }
 
+  const replaceAppName = useCallback(
```
lucasbento: What's the reason for using `useCallback`?

src/components/common/Settings.js
```diff
@@ -14,11 +14,24 @@ const Settings = ({ handleSettingsChange }) => {
     <Popover
       placement="bottomRight"
       content={
-        <Checkbox.Group onChange={updateCheckboxValues}>
-          <div>
-            <Checkbox value={SHOW_LATEST_RCS}>{SHOW_LATEST_RCS}</Checkbox>
-          </div>
-        </Checkbox.Group>
+        <Form>
```
lucasbento: Why do you need to use `Form`?

src/components/pages/Home.js
```diff
@@ -61,6 +61,7 @@ const Home = () => {
   const [settings, setSettings] = useState({
     [`${SHOW_LATEST_RCS}`]: false
   })
+  const [appName, setAppName] = useState('')
```
lucasbento: Is there a specific reason to not use `null` as the initial state?

Naturalclar: Personally, I prefer using empty string over `null`, as `useState` infers that `appName` will be a typeof `string`, and it's (imo) it's clear to the 
 reader that this will be a string.
It's also to keep code consistency as other places that uses `useState` for string seems to initiate the value with an empty string as well.
If I were to write this in typescript with `null`, I'd write it

```
const [appName, setAppName] = useState<string | null >(null)
```

and there'd have to be a null check all over the place, which makes the code harder to read (imo).

It's not too big of a deal and I'm fine with using `null` for initialization.

src/components/common/DiffViewer.js
Naturalclar: The current lint setup automatically adds `replaceAppName` as dependency of `useEffect` in DiffViewer.js
Since a function declaration is a different instance on every render, using `replaceAppName` without using `useCallback` to memoize the function will result in an infinite loop.

src/components/common/Settings.js
Naturalclar: I was looking for a good way to style the contents inside the settings, and Form seemed to have what the list-esque style that I was looking for.
It's not a requirement to use Form.

Naturalclar:
@pvinis @lucasbento 
Thanks so much for the feedback!
I've adjusted the style of the input a bit, and fixed the problem for typing fast.
the setTimeout seem to have done the trick! So awesome!

Kapture 2020-01-03 at 0 32 18

src/components/pages/Home.js
pvinis: I also prefer null but I guess for now `''` is fine for the reasons you mentioned. Maybe part of the rework should be typescript too? 🤔 Then we could do the template thing you mentioned. I'll add it.

lucasbento: > Personally, I prefer using empty string over null, as useState infers that appName will be a typeof string, and it's (imo) it's clear to the
reader that this will be a string.

I like this 🙂

kelset:
This looks so great ❤️ thanks @Naturalclar 👏

PR #126, opened 2019-12-30
Upgrades

pvinis:
# Summary

## Checklist


src/components/common/DiffViewer.js
```diff
@@ -69,9 +69,8 @@ const DiffViewer = ({
     const fetchDiff = async () => {
       setLoading(true)
 
-      const response = await (await fetch(
-        getDiffPatchURL({ fromVersion, toVersion })
-      )).text()
+      let response = await fetch(getDiffPatchURL({ fromVersion, toVersion }))
+      response = response.text()
```
lucasbento: Why did you need to do this?

pvinis: I actually changed it to simplify the nestesed awaits, and I forgot to add the await on the second line. this is reverted on master by the fix PR from Nat.

PR #129, opened 2019-12-31
Fix Failing Build

Naturalclar:
# Summary

Reverted some changes from #126 

Currently, developing in `master` branch causes the following error, when trying to show the diff.

Screen Shot 2019-12-31 at 1 27 51 PM

It seems to be caused from the lack of await in response?
Reverting the change made in #126 made the error go away.

Also, I'm getting error with `semver` after upgrading to `7.x`
Reverting the package to `6.3.0` seemed to fix the build.
There may be some breaking change in the `semver` package.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "react-ga": "2.7.0",
     "react-github-btn": "1.1.1",
     "react-scripts": "3.0.1",
-    "semver": "7.1.0",
+    "semver": "6.3.0",
     "styled-components": "4.4.1"
   },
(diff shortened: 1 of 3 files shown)


pvinis:
oooops, I forgot the await on the .text() line 🤦‍♂️🤦‍♂️🤦‍♂️

I prefer it in 2 lines instead of a nested await, but let's fix it here as you have it.

for semver I was trying to get rid of that error, I couldn't and had to leave.

thanks for this fix PR! 🙌

PR #156, opened 2020-01-22
[WIP] Reworking release filtering

pvinis:
# Summary
The release filtering has been a bit problematic for a while now. I am reworking it while adding tests, to make sure we don't break it again.

## Checklist


kelset:
(this is very old, probably not worth keeping around - if i'm wrong just open a new version from top of latest main)

PR #193, opened 2020-03-12
Add button to copy path to clipboard

benomatis:
# Summary
Adding a button to copy the file path to clipboard makes it easier to search for the file (eg. Command + Shift + O in JetBrain's IDE's) and make the necessary modifications. This is similar to the feature GitHub has.

Note that I'm using the `oldPath` going out from the assumption people will want to search for the file in the old version, and then do the modification to that.

1. Click the copy button next to the file name.
2. Paste clipboard content: you will see the file path added.
screencast 2020-03-12 18-29-02

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import styled from '@emotion/styled'
 import {
   Diff as RDiff,
@@ -95,6 +95,19 @@ const Diff = ({
   const [isDiffCollapsed, setIsDiffCollapsed] = useState(
     isDiffCollapsedByDefault({ type, hunks })
   )
+  const [isPathCopiedToClipboard, setIsPathCopiedToClipboard] = useState(false)
+
+  const handleCopyPathToClipboard = () => {
+    setIsPathCopiedToClipboard(true)
+  }
+
+  useEffect(() => {
+    if (isPathCopiedToClipboard === true) {
+      setTimeout(() => {
+        setIsPathCopiedToClipboard(false)
+      }, 5000)
+    }
+  }, [isPathCopiedToClipboard])
 
   if (areAllCollapsed !== undefined && areAllCollapsed !== isDiffCollapsed) {
     setIsDiffCollapsed(areAllCollapsed)
@@ -121,6 +134,8 @@ const Diff = ({
           setIsDiffCollapsed(collapse)
         }}
         isDiffCompleted={isDiffCompleted}
+        onCopyPathToClipboard={handleCopyPathToClipboard}
+        isPathCopiedToClipboard={isPathCopiedToClipboard}
         onCompleteDiff={onCompleteDiff}
         appName={appName}
       />
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useState } from 'react'
 import styled from '@emotion/styled'
 import {
   Diff as RDiff,
@@ -95,19 +95,25 @@ const Diff = ({
   const [isDiffCollapsed, setIsDiffCollapsed] = useState(
     isDiffCollapsedByDefault({ type, hunks })
   )
-  const [isPathCopiedToClipboard, setIsPathCopiedToClipboard] = useState(false)
+  const copyPathPopoverContentOpts = {
+    default: 'Copy path to clipboard',
+    copied: 'Copied!'
+  }
+  const [copyPathPopoverContent, setCopyPathPopoverContent] = useState(
+    copyPathPopoverContentOpts.default
+  )
 
   const handleCopyPathToClipboard = () => {
-    setIsPathCopiedToClipboard(true)
+    setCopyPathPopoverContent(copyPathPopoverContentOpts.copied)
   }
 
-  useEffect(() => {
-    if (isPathCopiedToClipboard === true) {
-      setTimeout(() => {
-        setIsPathCopiedToClipboard(false)
-      }, 5000)
+  const handleResetCopyPathPopoverContent = () => {
+    console.log('outside', copyPathPopoverContent)
+    if (copyPathPopoverContent === copyPathPopoverContentOpts.copied) {
+      console.log('inside', copyPathPopoverContent)
+      setCopyPathPopoverContent(copyPathPopoverContentOpts.default)
     }
-  }, [isPathCopiedToClipboard])
+  }
 
   if (areAllCollapsed !== undefined && areAllCollapsed !== isDiffCollapsed) {
     setIsDiffCollapsed(areAllCollapsed)
@@ -135,7 +141,8 @@ const Diff = ({
         }}
         isDiffCompleted={isDiffCompleted}
         onCopyPathToClipboard={handleCopyPathToClipboard}
-        isPathCopiedToClipboard={isPathCopiedToClipboard}
+        copyPathPopoverContent={copyPathPopoverContent}
+        resetCopyPathPopoverContent={handleResetCopyPathPopoverContent}
         onCompleteDiff={onCompleteDiff}
         appName={appName}
       />
--- a/src/components/common/Diff/DiffHeader.js
+++ b/src/components/common/Diff/DiffHeader.js
@@ -1,13 +1,15 @@
-import React, { Fragment } from 'react'
+import React, { Fragment, useEffect } from 'react'
 import styled from '@emotion/styled'
 import { Tag, Button, Popover } from 'antd'
 import {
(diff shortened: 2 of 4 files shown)


benomatis:
Let me review and correct the errors.

benomatis:
Should be complete now, waiting for your approval ;)

src/components/common/Diff/Diff.js
```diff
@@ -95,6 +95,19 @@ const Diff = ({
...
+
+  const handleCopyPathToClipboard = () => {
+    setIsPathCopiedToClipboard(true)
+  }
+
+  useEffect(() => {
+    if (isPathCopiedToClipboard === true) {
+      setTimeout(() => {
+        setIsPathCopiedToClipboard(false)
+      }, 5000)
```
pvinis: Do we unset the clipboard after 5 seconds?

benomatis: No, it's not effectively unsetting the clipboard, [it just changes the `content` of the `Popover`](https://github.com/react-native-community/upgrade-helper/pull/193/files/5d56cfe34f9e01ae7e6217a226e631130e493bc3#diff-6cc048559c1f2b450387cc327c82993cR148-R151):

```jsx
<Popover
    content={isPathCopiedToClipboard ? 'Copied!' : 'Copy path to clipboard'}
    trigger="hover"
>
```

benomatis:
> This is awesome, thank you so much for doing this!
> 
> Just one remark, do you know why the circle around the button shows like this on click? It looks like it has more `width` rather than `height`.
> Screenshot
> 
> I would say we just need to fix that and then it should be good to go 🙂

I copied the CSS from `CompleteDiffButton`, the reason you cannot see it there is because it disappears immediately. I'll fix it on the other one too to stay consistent, if that's ok with you.

lucasbento:
> @benomatis: I copied the CSS from CompleteDiffButton, the reason you cannot see it there is because it disappears immediately. I'll fix it on the other one too to stay consistent, if that's ok with you.

Yeah, sounds good, as long as the the circle around the button shows in a correct shape I'm good with it :)

benomatis:
> > @benomatis: I copied the CSS from CompleteDiffButton, the reason you cannot see it there is because it disappears immediately. I'll fix it on the other one too to stay consistent, if that's ok with you.
> 
> Yeah, sounds good, as long as the the circle around the button shows in a correct shape I'm good with it :)

Turns out it's the `shape` param that's set to `circle`. According to [the doc](https://ant.design/components/button/) there are 2 options, either `circle` or `round`, I tried `round` but that makes it even wider, if I omit it, it makes it square. I'll play with the css a bit to make it really a proper circle, as it seems that's what you're after.

benomatis:
Ok, I just did that. I collected some default CSS rules shared between the 2 buttons (`CompleteDiffButton` and this new one) so they can be set in a more consistent manner.

benomatis:
My IDE is acting up - I don't know why it suddenly changed one of the [`Button`'s to go onto one line](https://github.com/react-native-community/upgrade-helper/pull/193/commits/762ee0952975271097fb60fcdecf4e4bb6109fea#diff-6cc048559c1f2b450387cc327c82993cR155) instead of multiple. I will go ahead and change that to stay consistent, sorry about that.

```jsx
<Button {...props} type="ghost" icon={<CopyOutlined />} />
```

src/components/common/Diff/DiffHeader.js
```diff
@@ -149,22 +152,12 @@ const CopyPathToClipboardButton = styled(
         content={isPathCopiedToClipboard ? 'Copied!' : 'Copy path to clipboard'}
         trigger="hover"
       >
-        <Button
-          {...props}
-          type="ghost"
-          shape="circle"
-          icon={<CopyOutlined />}
-        />
+        <Button {...props} type="ghost" icon={<CopyOutlined />} />
```
benomatis: Will change this one to separate props into multiline

src/components/common/Diff/Diff.js
pvinis: I would prefer it if it didn't change.

- hover -> copy to clipboard
- click -> copied
- wait there -> copied
- move mouse away
- hover again -> copy to clipboard

Does that make sense?

src/components/common/Diff/DiffHeader.js
```diff
@@ -146,13 +146,22 @@ const CompleteDiffButton = styled(({ visible, onClick, ...props }) =>
...
+    resetCopyPathPopoverContent,
+    ...props
+  }) => (
     <CopyToClipboard text={removeAppPathPrefix(path, appName)} onCopy={onCopy}>
-      <Popover
-        content={isPathCopiedToClipboard ? 'Copied!' : 'Copy path to clipboard'}
-        trigger="hover"
-      >
-        <Button {...props} type="ghost" icon={<CopyOutlined />} />
+      <Popover content={copyPathPopoverContent} trigger="hover">
```
benomatis: This jumping to 1 line keeps happening; I tested, it's an IDE issue, even though when I do the committing it's not set to automatically change. I'll fix this, apologies.

src/components/common/Diff/Diff.js
benomatis: I think I addresssed what you wanted:
screencast 2020-03-13 13-22-34

src/components/common/Diff/DiffHeader.js
benomatis: Looks like prettier actually wants me to have those on one line, which is what my IDE is realising? Is it the built-in linter maybe...? Your cicleCI failed with prettier actually wanting me these on one line. What is your strategy on this @pvinis @lucasbento ?

benomatis: ...strange because all of my linters in my IDE (PHPStorm, btw) are disabled.

benomatis: ...ok, I figured it out: it's actually an awesome feature :) there is an option called "Run Git hooks" which does this, so it's actually a good thing. I'll commit again using it; promise this is the final one, sorry for the many comments.

pvinis: I look forward to prettier being banned.

src/components/common/Diff/Diff.js
pvinis: Awesome, thank you!

pvinis:
I think we are good to go here. Lucas' concern about the circle, my concern about the text, and (I'm assuming the name 😅) Ben's concern about formatting are all solved!

Thanks for the PR! 🙌

src/components/common/Diff/DiffHeader.js
benomatis: For the future, are you saying I should skip it instead? I thought that it was part of the git repo and that it was set up by you guys. It's supposed to follow what's been set up there - is that not something you are allowed to edit / modify as admins?

pvinis:
Sooo, it turns out I got too trigger-happy and merged early!
There are two `console.log`s that could be removed, and @lucasbento found a bug 
Kapture_2020-03-13_at_14 48 53

@benomatis Could you make these changes in a new PR?

Sorry to both of you for the inconvenience.

benomatis:
... and I hoped I won't have to fiddle around with CSS. Until now, what I found out is that the attribute `ant-click-animating-without-extra-node` seems to set itself to `true` randomly, and when it does, you get the `Copied!` text jumping. I'll try to fix it and create a new PR.

lucasbento:
> ... and I hoped I won't have to fiddle around with CSS. Until now, what I found out is that the attribute ant-click-animating-without-extra-node seems to set itself to true randomly, and when it does, you get the Copied! text jumping. I'll try to fix it and create a new PR.

@benomatis: thank you! I think this happens mostly when you click on it before the popover shows, perhaps you can try to check if it was clicked before it shows the popover so the content doesn't resize.

benomatis:
> @benomatis: thank you! I think this happens mostly when you click on it before the popover shows, perhaps you can try to check if it was clicked before it shows the popover so the content doesn't resize.

Yes, that seems to be the issue: it happens if you click too fast, before the default text fully shows.

benomatis:
Would this be too "quick and dirty"? I basically fixed the width of the tooltip - I'd understand why you would not prefer this option, just checking... :)

screencast 2020-03-13 16-12-25

lucasbento:
I honestly wouldn't mind that, @pvinis any thoughts on this?

pvinis:
What's the difference? 😬 We just don't have the blue bubble? I'm fine with that, yea.

benomatis:
@pvinis The difference is a slightly larger left and right padding when the "File path copied!" text is showing. Sorry, what blue bubble? :)

pvinis:
Oh, cool. Never mind, all good, I'm just blind, I didn't notice the difference! If it works then great. Make a new PR when you can, and we can try it ourselves there! :)

benomatis:
See https://github.com/react-native-community/upgrade-helper/pull/200

PR #200, opened 2020-03-13
Fix copy path popover content jumping

benomatis:
When you click the Copy path to clipboard button too quick, the `Copied!` text jumps to the left. I have fixed this by hard-coding the width of the popover.

This was the issue:
Kapture_2020-03-13_at_14 48 53

And this is how it looks after the change:

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -97,7 +97,7 @@ const Diff = ({
   )
   const copyPathPopoverContentOpts = {
     default: 'Copy path to clipboard',
-    copied: 'Copied!'
+    copied: 'File path copied!'
   }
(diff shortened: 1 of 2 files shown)


benomatis:
@pvinis Looks like you left `revert-193-copy-path-to-clipboard` there, was that on purpose...? I think you tried to revert the initial PR, then changed your mind and decided to add another one (this one) to fix the issue there was, but in the process you forgot about the revert branch, did I get it right?

pvinis:
You mean the branch? I'll delete it. Yea, basically that. I merged the other PR early, then I thought I'll revert and we do again, but in the end I thought it was easier to keep it and just make a new PR to finish the first.

PR #206, opened 2020-03-18
Keep original path with `appName` of `RnDiffApp` to View or Download files

benomatis:
If you change the App Name on top, then click on View file or the download icon, you'll get redirected to a dead link because the app name is also replaced in the path to the one the user set.

I changed it so it reverts to the default RnDiffApp only for view and download to prevent the 404 error.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -15,6 +15,9 @@ export const getBinaryFileURL = ({ version, path }) =>
 export const removeAppPathPrefix = (path, appName) =>
   path.replace(new RegExp(`${appName || 'RnDiffApp'}/`), '')
 
+export const getOriginalPath = (path, appName) =>
+  path.replace(new RegExp(`${appName}`), 'RnDiffApp')
+
 export const getVersionsInDiff = ({ fromVersion, toVersion }) => {
   const cleanedToVersion = semver.valid(semver.coerce(toVersion))
 
--- a/src/components/common/Diff/DiffHeader.js
+++ b/src/components/common/Diff/DiffHeader.js
@@ -8,7 +8,11 @@ import {
   RightOutlined,
   CopyOutlined
 } from '@ant-design/icons'
-import { removeAppPathPrefix, getBinaryFileURL } from '../../../utils'
+import {
+  removeAppPathPrefix,
+  getBinaryFileURL,
+  getOriginalPath
+} from '../../../utils'
 import { CopyToClipboard } from 'react-copy-to-clipboard'
 
 const Wrapper = styled.div`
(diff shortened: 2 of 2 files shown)


pvinis:
Good catch. I think we should do some work there for actually supporting this better, but this is not for this PR.

lucasbento:
> @pvinis: Good catch. I think we should do some work there for actually supporting this better, but this is not for this PR.

Would be nice to create an issue for this if you already have ideas in mind 😄

pvinis:
Let's merge this for now, because it's better to show the file than to show an error 😅

benomatis:
> > @pvinis: Good catch. I think we should do some work there for actually supporting this better, but this is not for this PR.
> 
> Would be nice to create an issue for this if you already have ideas in mind 😄

I guess this will have to be done with apache mod_rewrite, to keep the URL with the custom app name in the path, but show the content of the file with the original path.

pvinis:
My idea was to not show the original, basically #208.

src/utils.js
```diff
@@ -15,6 +15,9 @@ export const getBinaryFileURL = ({ version, path }) =>
 export const removeAppPathPrefix = (path, appName) =>
   path.replace(new RegExp(`${appName || 'RnDiffApp'}/`), '')
 
+export const getOriginalPath = (path, appName) =>
+  path.replace(new RegExp(`${appName}`), 'RnDiffApp')
```
benomatis: @pvinis Following this change, for some reason, ios paths show the app name twice at the beginning resulting in a 404 again.

eg https://github.com/react-native-community/rn-diff-purge/raw/release/0.59.1/RnDiffAppRnDiffApp/package.json

notice `RnDiffAppRnDiffApp` in the path just after the version number.

I'm wondering it it has something to do with my implementation of the `replace` method.

Should it maybe just be this...?

```suggestion
  path.replace(appName, 'RnDiffApp')
```

benomatis: I tried this, it's not that... I'll try to fix this and open a new pull request, if you have any idea, I'd welcome any pointers ;) Thanks.

benomatis: in fact, this seems to be a problem with files that have the app name in it (like some of the ios ones)...

benomatis: Should be fixed in https://github.com/react-native-community/upgrade-helper/pull/209

elie222:
Possibly broke this:
https://react-native-community.github.io/upgrade-helper/?from=0.60.6&to=0.61.5

benomatis:
> Possibly broke this:
> https://react-native-community.github.io/upgrade-helper/?from=0.60.6&to=0.61.5

Looks like it's broken by eeea806ed25e9e1cbca66f5c2f1dd0f1d907163a instead.

lucasbento:
Thanks for notifying this issue @elie222 and thank you @benomatis for pinpoint where the problem was.

I just reverted the version bump in https://github.com/react-native-community/upgrade-helper/commit/eeea806ed25e9e1cbca66f5c2f1dd0f1d907163a so the latest build should be working.

benomatis:
@lucasbento Yup, I can confirm, it works with the old version.
@elie222 All good for you too?

elie222:
Yup, thanks!
https://react-native-community.github.io/upgrade-helper/?from=0.60.6&to=0.61.5 works great

PR #209, opened 2020-03-18
Fix file path when viewing or downloading file

benomatis:
Related to https://github.com/react-native-community/upgrade-helper/pull/206, I unfortunately used an incorrect Regex, it's fixed here. Apologies.

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -78,6 +78,7 @@ const isDiffCollapsedByDefault = ({ type, hunks }) =>
 const Diff = ({
   oldPath,
   newPath,
+  originalPath,
   type,
   hunks,
   fromVersion,
@@ -124,6 +125,7 @@ const Diff = ({
       <DiffHeader
         oldPath={oldPath}
         newPath={newPath}
+        originalPath={originalPath}
         toVersion={toVersion}
         type={type}
         diffKey={diffKey}
--- a/src/components/common/Diff/DiffHeader.js
+++ b/src/components/common/Diff/DiffHeader.js
@@ -9,11 +9,7 @@ import {
   CopyOutlined,
   RollbackOutlined
 } from '@ant-design/icons'
-import {
-  removeAppPathPrefix,
-  getBinaryFileURL,
-  getOriginalPath
-} from '../../../utils'
+import { removeAppPathPrefix, getBinaryFileURL } from '../../../utils'
 import { CopyToClipboard } from 'react-copy-to-clipboard'
 
 const Wrapper = styled.div`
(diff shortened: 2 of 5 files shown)


lucasbento:
Hey @benomatis, thank you for fixing this.

There's however still an issue, if you go to https://deploy-preview-209--upgrade-helper.netlify.com/?from=0.61.5&to=0.62.0-rc.5, then change the name of the app to something like `awesomeapp`, look for a file like `android/app/src/debug/java/com/awesomeapp/ReactNativeFlipper.java` and then tap `View File` you will see that it goes 404.

That's because it should lead to `android/app/src/debug/java/com/rndiffapp/ReactNativeFlipper.java` instead.

Would you have time to look into that?

benomatis:
> Hey @benomatis, thank you for fixing this.
> 
> There's however still an issue, if you go to https://deploy-preview-209--upgrade-helper.netlify.com/?from=0.61.5&to=0.62.0-rc.5, then change the name of the app to something like `awesomeapp`, look for a file like `android/app/src/debug/java/com/awesomeapp/ReactNativeFlipper.java` and then tap `View File` you will see that it goes 404.
> 
> That's because it should lead to `android/app/src/debug/java/com/rndiffapp/ReactNativeFlipper.java` instead.
> 
> Would you have time to look into that?

I guess the best approach would be to store the original paths, even upon app name change, and use them for view and download; the regex could become quite complex if we wanted to account for these.

lucasbento:
> I guess the best approach would be to store the original paths, even upon app name change, and use them for view and download; the regex could become quite complex if we wanted to account for these.

Yeah, I completely agree with this.

benomatis:
@lucasbento It took me a while because the `appName` is changed in `DiffViewer` from the diff before it being parsed, but it can only be passed on to `DiffSection` after having been parsed, so I ended up parallelly parsing the diff but not applying the `replaceAppName` on it, and using the index of the diff with unchanged app name to get it's `newPath` and supply it to the `originalPath` variable. I hope this is not to complex or resource hungry, I don't see a performance degradation, the way I implemented it, but feel free to simplify, of course.

lucasbento:
Hey @benomatis, thanks for changing it.

I'm gonna take a better look at the diff later today, perhaps we don't need to parse the diff twice to get it fixed, I'm a little bit worried about how much the processing is increased by doig that.

benomatis:
> Hey @benomatis, thanks for changing it.
> 
> I'm gonna take a better look at the diff later today, perhaps we don't need to parse the diff twice to get it fixed, I'm a little bit worried about how much the processing is increased by doig that.

Sure, please do, I still have lots to learn ;)

benomatis:
Happy for this to be closed for https://github.com/react-native-community/upgrade-helper/pull/214

PR #213, opened 2020-03-19
Revert "Keep original path with `appName` of `RnDiffApp` to View or Download files"

lucasbento:
Reverts react-native-community/upgrade-helper#206


benomatis:
@lucasbento Try checking out my previous changes, see there is no error, then check out eeea806 and the error is there. Looks like there is something in the  `"react-diff-view": "2.4.3"` patch version causing this.

lucasbento:
@benomatis looking into it

Issue #208, opened 2020-03-18
View/Download file with changed app name


pvinis:
## Description

Let's take `package.json` as an example.

When I change the app name from the options popover, I can see that in the diff viewer the `"name"` is `"myApp"` for example, or whatever I put in the options.

When I click to view or download a file, the `"name"` in there is `"RnAppDiff"`, which makes sense, since it's just the original file.

## Discussion

I don't know if that is confusing or possibly causing trouble. I guess we could fix this by displaying the raw file after we have replaced the app name, like we do for the diff viewer, but is it worth it?

benomatis:
Wouldn't you actually need to create the page with the app name in order to be able to display the contents of the original path? Alternatively we could do this with apache's mod_rewrite, then again that could become a quite complicated script.

And don't forget to think of not just the folder names in the path, but some file names have the app name as well, like many iOS files.

pvinis:
Let's close this. This is not worth our time, not right now at least. There are more important things we can do in this repo and the upgrade story. :)

Maybe we can revisit in the future.

PR #214, opened 2020-03-19
Use app name replacement only in the file header and not the content

lucasbento:
# Summary

We have noticed a good amount of bugs in the last few days because of the `replace`s we do in order to change `RnDiffApp` into the name that the user chooses in the interface.

This PR changes the approach of doing this from replacing all the diff changes to only doing so for the file header and the copy file path functionality which kills the bugs we have been experiencing with the `View File` link and download button.

## Test Plan

1. Put in a few version combinations;
1. Change the app name;
1. Check that:
	- The file header is correct;
	- The `View File` goes to the right place;
	- The copy button copies the correct path with the correct app name.

Do the same without changing the app name.

### DIFF
--- a/src/components/common/Diff/Diff.js
+++ b/src/components/common/Diff/Diff.js
@@ -96,7 +96,7 @@ const Diff = ({
     isDiffCollapsedByDefault({ type, hunks })
   )
   const copyPathPopoverContentOpts = {
-    default: 'Copy path to clipboard',
+    default: 'Click to copy file path',
     copied: 'File path copied!'
   }
   const [copyPathPopoverContent, setCopyPathPopoverContent] = useState(
--- a/src/components/common/Diff/DiffHeader.js
+++ b/src/components/common/Diff/DiffHeader.js
@@ -12,13 +12,23 @@ import {
 import {
   removeAppPathPrefix,
   getBinaryFileURL,
-  getOriginalPath
+  getPathWithProvidedAppName
 } from '../../../utils'
 import { CopyToClipboard } from 'react-copy-to-clipboard'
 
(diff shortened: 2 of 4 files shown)


lucasbento:
@benomatis can you please take a look? let me know your thoughts!

Unfortunately this supersedes the work you have done in #206.

lucasbento:
Also, it's good to keep in mind that with this PR we will keep showing stuff like `com.rndiffapp` in the diff as we will be no longer replacing that.

Given the number of bugs we had related to this I would rather proceed with it without this functionality as it's not that useful IMO.

pvinis:
While we're here, I'll go ahead and close #208. I don't think it's worth it. We already have a notice above the diffs that the name `RnDiffApp` is a template thing.

benomatis:
@lucasbento Looks good, nice and simple ;) ... Though what is remaining that would need the app name in it? Isn't it the `hunk` only? Anything else? If it's the only one, it wouldn't be complicated to just run that one through a (or the same) replace function, no?

lucasbento:
> @benomatis Looks good, nice and simple ;) ... Though what is remaining that would need the app name in it? Isn't it the hunk only? Anything else? If it's the only one, it wouldn't be complicated to just run that one through a (or the same) replace function, no?

Indeed only the hunks, I looked into replacing the name there but I think this could cause issues performance-wise and I would rather not go down that rabbit hole.

lucasbento:
I'm gonna merge this (as it's current broken in `master`) but feel free to ask any questions or request any other implementation @benomatis & @pvinis.

Btw thank you for your continuous help around this, @benomatis!

PR #218, opened 2020-03-26
changing the anchor style

pvinis:
# Summary

The anchor style changed.

- `0<minor>`
https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#060
- `0<minor>0`https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#0610
- `v0<minor>0`https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0620

For now I just changed this to work for the latest release, which means it breaks the links to the other older releases. 🤷‍♂️

## Test Plan
Go to a diff from 61 to 62, and click the changelog link

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -43,11 +43,10 @@ export const getVersionsInDiff = ({ fromVersion, toVersion }) => {
   })
 }
 
+const baseChangelogURL =
+  'https://github.com/react-native-community/releases/blob/master/CHANGELOG.md'
 export const getChangelogURL = ({ version }) =>
-  `https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#${version.replace(
-    '.',
-    ''
-  )}`
+  `${baseChangelogURL}#v${version.replace('.', '')}0`
 
 // settings constants
 export const SHOW_LATEST_RCS = 'Show latest release candidates'


src/utils.js
```diff
@@ -43,11 +43,10 @@ export const getVersionsInDiff = ({ fromVersion, toVersion }) => {
...
 }
 
+const baseChangelogURL =
+  'https://github.com/react-native-community/releases/blob/master/CHANGELOG.md'
 export const getChangelogURL = ({ version }) =>
-  `https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#${version.replace(
-    '.',
-    ''
-  )}`
+  `${baseChangelogURL}#v${version.replace('.', '')}0`
```
lucasbento: This won't really work for previous versions, right?
I think we need to change `0.62.0` to remove the `v` and then change the other tags in the changelog to include the patch version.

pvinis: Let's keep this PR and ping @alloy and @grabbou to see if we will change the https://github.com/react-native-community/releases/blob/master/CHANGELOG.md file and make it consistent, and then we can finalize this PR.

alloy: I’m in favor of going back and fix the CHANGELOG 👍

alloy: How would this change work for patch releases of RN? Or does this tool not support patch releases?

pvinis: For now, we don't link to patch releases. Supposedly this link is for the "big" changes between releases.

But the change would be easy. Instead of the last hardcoded `0` there, we just put the patch number.

lucasbento:
PR for fixing the changes: https://github.com/react-native-community/releases/pull/181
Let's wait for that to be approved them we can go ahead and merge this :D

PR #219, opened 2020-03-27
Hide gradle.bat file change from 0.61 to 0.62 & Add a few links to 0.62

lucasbento:
# Summary

This PR hides the file change of `gradle.bat` as the changes are not really useful and adds a few links to `0.62` release.

### DIFF
--- a/src/releases/0.62.js
+++ b/src/releases/0.62.js
@@ -4,24 +4,32 @@ export default {
   usefulContent: {
     description:
       'React Native 0.62 includes built-in integration with Flipper.',
-    links: []
-  },
-  comments: [
-    {
-      fileName: '.gitignore',
-      lineNumber: 23,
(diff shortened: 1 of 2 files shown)


src/releases/0.62.js
```diff
@@ -4,24 +4,32 @@ export default {
...
+      {
+        title:
+          '[iOS] Tutorial on upgrading Xcode-related files to React Native 0.62',
+        url:
+          'https://github.com/react-native-community/upgrade-helper/issues/191'
+      }
+    ],
+    comments: [
+      {
+        fileName: 'ios/RnDiffApp.xcodeproj/project.pbxproj',
```
lucasbento: @pvinis is this in the correct place? I was wondering if we should put it here or in the `.xcodeproj`.

PR #225, opened 2020-03-30
Add Upgrade Support link to `UsefulContentSection`

lucasbento:
# Summary

This PR adds a mention to the [Upgrade Support](https://github.com/react-native-community/upgrade-support) repository at the top inside of the `Useful Content Section`.

---

Looks like this:

image

And also has a tooltip when hovering:

image


lucasbento:
@kelset: tagging you in as I would appreciate your input, your writing is :chef-kiss: 😄

PR #246, opened 2020-06-03
[WIP] Troubleshooting section

lucasbento:
# Summary

⚠️ STILL WIP ⚠️

This adds a section to Upgrade-Helper that _will_ bring valuable issues from Upgrade-Support to the attention of the users that are upgrading.

The way that it will work is that it will rely on a specific label to fetch all the issues related to the version range selected.

## TODOs

- [x] Separate common components from `UsefulContentSection`
- [x] Should have the same _look_ and _feel_ of `UsefulContentSection`
- [x] Add pagination (limit to 5)
- [x] Show labels to identify "Questions" or "Solutions"
- [ ] Fetch from [GitHub API](https://developer.github.com/v3/issues/#list-repository-issues)
- [x] Add a color less _extravagant_ to the section (previously using yellow, now gray-ish)
- [ ] Only gather issues between the version range specified
- [ ] Should be initially closed (needs to be discussed)
- [ ] Add some kind of animation in between the pagination

## Showcase

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@emotion/core": "10.0.28",
     "@emotion/styled": "10.0.27",
     "antd": "4.0.3",
-    "framer-motion": "^2.0.0-beta.52",
+    "framer-motion": "^2.0.0-beta.57",
     "markdown-to-jsx": "6.11.0",
     "query-string": "6.11.1",
     "react": "16.13.0",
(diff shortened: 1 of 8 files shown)


kelset:
(this is very old, probably not worth keeping around - if i'm wrong just open a new version from top of latest main)

PR #249, opened 2020-07-30
remove unused prop in Diff Header Component

luism3861:
# Summary
this PR remove unused prop in FileName arrow function!!.


kelset:
Hey folks — I’m doing a maintenance pass on the open PRs and finally catching up on some of the very old ones. Sorry this one sat for so long.

I’m going to close this PR because it’s effectively obsolete at this point. The cleanup itself is tiny and sensible, but it targeted the old JS version of `DiffHeader`, and the codebase has moved on substantially since then. In the current TypeScript version of the file, this specific unused-prop cleanup is no longer relevant as an open change.

Thanks again for the cleanup here — I just don’t think there’s anything left to merge from this branch anymore.

PR #265, opened 2021-05-06
Fix hermes -> Hermes and optin-in -> opt-in

brentvatne:
# Summary

Fix the typo and use correct casing for Hermes (as we do for React Native, since we don't write react native here)


lucasbento (approved):
Thank you @brentvatne!

PR #276, opened 2021-06-29
Add image comparison tests

### LINKED ISSUE #53: Test automatically on different browsers
# Feature Request

We should test the website automatically on multiple browsers.

## Why it is needed

With #51 being opened (thank you for this @KevinHu2014), it came to my attention that we should properly test the website on multiple browsers, the best way to do it would be to run a few tests automatically on CircleCI.

## Possible implementation

~~We can use [Cypress.io](https://docs.cypress.io/guides/guides/continuous-integration.html#Basics) with CircleCI, we can also think about a few image snapshots to do image comparison, which is an easier way to get things going for now.~~

Apparently Cypress supports only Chrome, we may consider using Sauce Labs (where we can run on multiple browsers) and [WebdriverIO](https://webdriver.io/) together with [wdio-image-comparison-service](https://github.com/wswebcreation/wdio-image-comparison-service).

I recall seeing a tweet that [now](now.sh) also had a plugin/extension for doing image comparison on PRs, let's see if I find more info about that.

lucasbento:
# Summary

~~Kinda~~ fixes #53.

This PR adds [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) and [puppeteer](https://github.com/puppeteer/puppeteer) to automatically generate screenshots of different sections of the application and compare them using jest, it also disables the animations of a few elements if the browser is being controlled by puppeteer.

It's still WIP because it fails on CI even though it passes locally, the issue is that generated screenshots on macOS will fail as CI uses Ubuntu, the recommendation is that we generate the images on a docker container (see americanexpress/jest-image-snapshot#177) which needs to be implemented.

## Test Plan

- `yarn test` should pass locally and in CI.

## Checklist

- [x] Generate screenshots in a docker container to make sure they match with CI's


lucasbento:
# @pvinis IT'S FINALLY FINISHED

pvinis (approved):
im definitely afraid of this blocking PRs, but we manage that when we get there :D

PR #278, opened 2021-08-08
Add troubleshooting guides

lucasbento:
# Summary

This PR removes the current `Alert`s for the issues with Xcode 12.5 and Apple Silicon (M1) machine and adds a troubleshooting guide button.

I went a little bit over the top with the time invested on this but I think it looks great, I didn't do any handling for the versioning (yet) which should be fine for now.

The initial content (`Check here the troubleshooting guides for issues with Xcode 12.5 or Apple Silicon (M1) machine.`) will only show once to the user and, once it's closed, it will be shown only after two days.

I had to do some weird handling to know when to close or change the content because of the events that antd provides for `Popover` which are basically way too simple.

Troubleshooting Guides

---

cc @pvinis @kelset

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,8 @@ module.exports = {
   extends: 'react-app',
   plugins: ['prettier'],
   rules: {
-    'prettier/prettier': 'error'
+    'prettier/prettier': 'error',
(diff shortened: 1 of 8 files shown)


lucasbento:
Gonna merge this now since I discussed with @pvinis already but let me know if you have any opinions/ideas.

PR #291, opened 2021-10-31
Fix path not properly being shown without `RnDiffApp`

lucasbento:
# Summary

There was some sort of change when generating releases that caused the diff to change as following:

image

Which led `upgrade-helper` to show the header as:

image

It seems like the lib we use for parsing diff doesn't like the lack of those initial characters so I had to add some replaces here and there as well as maintain the `RnDiffApp` replace for the previous versions.

## Test Plan

- Open the dev preview;
- Make sure that the paths are rendering correctly without the leading `RnDiffApp` but still using in specific places (such as `RnDiffApp.xcodeproj`);
- Set a different app name;
- Make sure that it renders correctly with the different app name.

## Checklist

### DIFF
--- a/src/components/common/DiffViewer.js
+++ b/src/components/common/DiffViewer.js
@@ -130,7 +130,7 @@ const DiffViewer = ({
   }
 
   const diffSectionProps = {
-    diff: diff,
+    diff,
     getDiffKey: getDiffKey,
     completedDiffs: completedDiffs,
     fromVersion: fromVersion,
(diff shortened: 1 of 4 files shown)


kelset:
Hey folks — I’m doing a maintenance pass on the open PRs and finally catching up on some older ones. Sorry this one sat for so long.

I’m going to close this PR at this point. The branch is old enough that it predates a lot of changes in the repo, including the TypeScript migration and later fixes around app-name/path handling.

Because of that, I don’t think this is a realistic branch to salvage anymore:
- it targets older JS-era files that have since moved/changed,
- the code around path rendering has evolved quite a bit since 2021,
- and if there’s still any remaining edge-case bug here, the better move would be a fresh narrowly scoped PR from current `master`.

Thanks for the original fix attempt here — I just don’t think this specific branch is mergeable anymore in the current codebase.

PR #294, opened 2021-11-09
Add react-native-macos support and UI to switch packageName/Language

acoates-ms:
# Summary

Adds UI to the settings popover to allows users to switch between `react-native` and `react-native-windows`, and to switch between C++ and C# within `react-native-windows`.

Here is what the new settings popover looks like:
image

## Test Plan

Ran through the site starting with various values of `packageName` and `language` in the URL.  Try changing the package name back and forth in the settings.  Try changing the language when `packageName=react-native-windows`.  Try changing it when in `packageName=react-native`  Try changing back to `react-native` after switching the language.

Just load the site and hit the settings icon in the top right.

## Checklist

### DIFF
--- a/src/components/common/Settings.js
+++ b/src/components/common/Settings.js
@@ -50,19 +50,19 @@ const Settings = ({
       onChangeAppName(newAppName)
     }
 
+    const processedNewLanguage =
+      newLanguage !== language && newPackageName === PACKAGE_NAMES.RNW
+        ? newLanguage
+        : LANGUAGE_NAMES.CPP
+
     if (
       !visibility &&
-      (newPackageName !== packageName || newLanguage !== language)
+      (newPackageName !== packageName || processedNewLanguage !== language)
     ) {
       onChangePackageNameAndLanguage({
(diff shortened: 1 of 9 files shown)


lucasbento:
Hey @acoates-ms, this is an awesome addition to the Upgrade-Helper!

The initial load of the diff after the you change the package doesn't seem to work, it does work afterwards though:

https://user-images.githubusercontent.com/6207220/141144934-368796f3-7831-4906-be51-15790552e825.mp4

acoates-ms:
The reason for this is the to/from versions not being valid for the new platform.  So instead I'm going to reset the to/from versions and hide the diff rather than showing the broken diff.  That way it doesn't look as broken.  The user will still have to select the versions and hit `Show me how to upgrade!` button again to show the new diff.

acoates-ms:
@lucasbento should be good for another look.

lucasbento (approved):

lucasbento:
Looks great, thank you @acoates-ms!

PR #296, opened 2021-11-23
fix top section layout on mobile

edenstrom:
# Summary

Haven't created an issue. It fixes the mobile layout for the top section in the application, excluding the diff viewer.

## Test Plan

Changes the layout for devices below 768px width. Portrait iPads get the normal layout.

The only change visible on desktop devices is the gap between troubleshooting and settings which is `15px` instead of `8px`. 

No changes in dependencies.

Tested using iPhone 11 Pro and Chrome, Firefox, Safari on macOS 11.6.1.

The popper.js arrow is not in the correct position. May have to do some extra tweaks for that in a separate PR.

<table>
<tr>
	<th>Before</th>
	<th>After</th>
</tr>
<tr>
	<td>before</td>
	<td>after</td>
</tr>
<tr>
	<td>before2</td>
	<td>after2</td>
</tr>
</table>

1. `yarn start`
2. Check UI

## Checklist

### DIFF
--- a/src/components/common/TroubleshootingGuidesButton.js
+++ b/src/components/common/TroubleshootingGuidesButton.js
@@ -18,7 +18,6 @@ const Icon = styled.span`
 const Button = styled(AntdButton)`
   width: 32px;
   padding: 0;
-  margin-right: 8px;
   color: initial;
 `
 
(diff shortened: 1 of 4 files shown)


lucasbento (approved):
Thank you so much for fixing this!

PR #355, opened 2023-04-01
Try to improve appName setting UX

zoontek:
Hey 👋

A small PR to add something I always wanted to the react-native upgrade app: the possibility to copy new raw file content with the replaced appName in it.
Today, if we press the "View file" button, it opens a new tab with the raw file, but without our replaced appName. I'm using the GitHub API (public? I don't think there's a limitation for that).

I also made a few UI changes:

- Always show the app name input, as it's too important to hide inside the settings popup imho
- Replace "MyAwesomeApp" placeholder with the default, "RNDiffApp" to make setting goal easier to understand
- Fix buttons sizes inconsistency
- Add back `react-native-windows` text in the settings popup (currently not visible)
- Replace "View file" text with a "Raw" button, to fit Github style:

Screenshot 2023-04-01 at 16 15 59

Of course, you can ask for all the modifications / rollbacks you want 🙂

---

Screenshot 2023-04-01 at 16 22 33


zoontek:
A quick demo: https://twitter.com/zoontek/status/1642246861544124418?s=46&t=2wABsXbRq79OEsT0SgBmrQ

pvinis (approved):
nice, thanks!!

PR #359, opened 2023-04-10
Fix copy on Safari

zoontek:
Following https://github.com/react-native-community/upgrade-helper/pull/355, [one of my follower](https://twitter.com/alihdjr/status/1643956095575048192) noticed that it doesn't work correctly on Safari.

Screenshot 2023-04-10 at 12 00 27

This occurs because the copy to clipboard is not done immediately after user interaction, in a synchronous way (as we have to fetch the file content before).

This PR fixes that, following [this guide](https://wolfgangrittner.dev/how-to-use-clipboard-api-in-firefox/). Tested successfully on Chrome / Firefox / Safari. Support is IMHO [good enough](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).


pvinis (approved):
good catch. thanks!

PR #360, opened 2023-04-26
feat: Add changelog link

blakef:
Let users open a tab with the CHANGELOG, and just to the target version.

# Summary

Provides anchor links to the CHANGELOG for the target version.

CleanShot 2023-04-26 at 16 00 17@2x

This is more useful or vanilla React Native.  For windows and macos it link to their release tags.  Caveat: because these projects aren't always synced to the latest React Native releases, some of these links may not exist.

## Test Plan

Added some unit tests and tested on my device.

## Checklist


pvinis (approved):
nice thanks

PR #368, opened 2023-09-03
feat: add search option select from version and to version

fxamauri:
# Summary
Add search option in selectors to version and from version.

uhelper-search

## Test Plan
- focus on one of the selectors when typing something, it should suggest a list item

## Checklist


markrickert:
Took me a moment to realize you just added the prop to an `antd` [`Select`](https://ant.design/components/select#select-props) component. I like this change! 👍

lucasbento (approved):
Great finding @fxamauri! Thank you for working on this ❤️

PR #369, opened 2023-09-08
Fix app name / package replacement

zoontek:
# Summary

Currently, app name / package replacement is kinda broken.

Screenshot 2023-09-08 at 19 03 25

Let's say we have this scenario: My app name is `Swan`, my app package is `io.swan.id`. With this config, on the current version of the upgrade helper, you will notice that the app name is append to the app package, making it `io.swan.id.swan`

This is visible in paths or if you copy file raw content:

Screenshot 2023-09-08 at 19 42 51

Screenshot 2023-09-08 at 19 45 22

---

This PR fixes all of this. It also:
- Fixes the app package placeholder
- Fixes the tests (which are actually expecting an incorrect result, as we can also see that the app name is appended)
- Continues to use `com.${appName.toLowercase()}` if app name is set, but app package is not

## Test Plan

Check both files content and paths with:
- no app name + no app package
- an app name but no app package
- an app name and an app package

## Checklist

### DIFF
--- a/src/__tests__/utils.spec.js
+++ b/src/__tests__/utils.spec.js
@@ -96,7 +96,7 @@ describe('replaceAppDetails ', () => {
       'android/app/src/debug/java/com/rndiffapp/ReactNativeFlipper.java',
       'SuperApp',
       'au.org.mycorp',
-      'android/app/src/debug/java/au/org/mycorp/superapp/ReactNativeFlipper.java',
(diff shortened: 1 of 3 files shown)


pvinis:
can you try on the netlify preview to see if it works as you expect?

zoontek:
@pvinis Just checked it, works correctly ✅

pvinis (approved):
checked too. seems working. thank you for the PR!

PR #378, opened 2024-02-26
Fix duplicate hunks when using updated hunks list

blakef:
# Summary

I just took a stab at what I think is causing #377, it'd be useful to get some eyes to validate why we were originally mapping each list of `hunks` to `updatedHunks`.  Here's the current state:

CleanShot 2024-02-26 at 15 56 15@2x

After the change:

CleanShot 2024-02-26 at 16 05 45@2x

## Test Plan

`yarn start`

## Checklist

### DIFF
--- a/src/components/common/Diff/Diff.tsx
+++ b/src/components/common/Diff/Diff.tsx
@@ -272,7 +272,10 @@ const Diff = ({
     toVersion,
   })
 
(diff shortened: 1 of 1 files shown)


markrickert (approved):
I tested this locally on my machine and it looks good!

lunaleaps (approved):

lunaleaps:
Thanks for fixing @blakef!

PR #379, opened 2024-02-26
Fix issue with AppName stripping incorrectly.

blakef:
Handles cases where appName wasn't playing nicely with it's default value.

# Summary
The `appName` is meant to be stripped off in various places.  When setting the default, the actual value passed around is `''`.  This change does 2 things:

1. makes regex stricter to only match against prefixes, and
2. passes the default `appName` down the tree when an empty string is inputted.

## Before
CleanShot 2024-02-26 at 16 05 45@2x

## After

https://github.com/react-native-community/upgrade-helper/assets/49578/5b467ea7-2415-4879-a40b-4bf4cbeb701d

## Test Plan

`yarn start`

You can see it live.

## Checklist

### DIFF
--- a/src/__tests__/utils.spec.ts
+++ b/src/__tests__/utils.spec.ts
@@ -2,6 +2,7 @@ import { PACKAGE_NAMES } from '../constants'
 import '../releases/__mocks__/index'
 import {
   getVersionsContentInDiff,
+  removeAppPathPrefix,
   replaceAppDetails,
   getChangelogURL,
 } from '../utils'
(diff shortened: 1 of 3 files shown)


blakef:
Well that's note entirely working 😉... BRB.

frankcalise:
Should close #381 and #382

kelset:
Hey folks — I’m doing a maintenance pass on the open PRs and finally catching up on some older ones. Sorry this took us such a long time to come back around to.

I think this PR has effectively been superseded by later merged work in the same area, especially #396, which addressed the app-name/file-path behavior from current `master`.

Given that:
- this branch is now fairly old,
- the relevant code has moved on since it was opened,
- and the specific path-formatting issue has since been addressed elsewhere,

I don’t think it makes sense to try to merge or revive this branch as-is anymore.

So I’m going to close this one out as superseded by later work. Thanks again for taking a stab at this.

PR #389, opened 2024-10-01
Fix older versions of react native changelog URLs

ItaloFalzoni:
# Summary

The previous React Native changelog files have been changed to another file. For example, the version 0.68 of react native has been changed to this link: https://github.com/facebook/react-native/blob/main/CHANGELOG-pre-070.md#v0680, before we were use this link: https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v068 and it was not working very well.

Basically I added an extra function that checks versions prior to 0.70 and invokes a new function that is responsible for assembling the changelog URLs.

## Test Plan

`yarn start`

You can see it live and check the link on the "Useful content for upgrading" section

## Checklist


pvinis (approved):

JVGS1111:
Can you guys check this?

kelset:
@ItaloFalzoni can you verify that TS checks green locally? on CI it errors with `Cannot set property mark of #<Object> which has only a getter`

ItaloFalzoni:
@kelset

Sorry. I coded on Github Codespace and I didn't see how the TS check works. As soon as possible I will look again to make the review easier. Thanks.

kelset (changes requested):
Thanks for putting this together — and sorry it took us such a long time to come back around to this one.

The underlying issue still makes sense, but I don’t think we can merge this PR as-is anymore because the React Native changelog layout has changed again since this was opened.

The changelog files we need to account for now are:

- `CHANGELOG-0.5x.md`
- `CHANGELOG-0.6x.md`
- `CHANGELOG-0.7x.md`
- and `CHANGELOG.md` for newer releases

So if you’re still interested in updating this PR, could you please:

- update/rebase the branch on current `master`
- update the changelog URL logic to use the current file layout above
- ideally use semver-based comparisons rather than string comparisons for version routing
- add/update tests around `getChangelogURL()` for representative versions

If you’re still up for refreshing it, I’m happy to take another look after that. And if not, no worries at all — we can close this one later and revisit with a fresh PR from current `master`.

PR #396, opened 2025-02-27
fix: correctly replace app name in file path

vinzscam:
# Summary

Fixed an issue causing file paths to lose the first slash when using `RnDiffApp` as app name or an empty string after editing the input. A similar fix might need to be applied to `appPackage`, but I refrained from doing so to avoid potential side effects on any underlying component, as the value can be nullable.

Screenshot 2025-02-27 at 18 10 05

I know there was already #379 open, but we used this fix in [our fork](https://github.com/backstage/upgrade-helper) and we thought it was good to contribute upstream. Feel free to ignore this PR if another fix is more suitable 😊

1. Type: `RnDiffApp` in the app name input
2. Press "Show me how to upgrade!"
3. You will see that the path of each file is wrong.

or

1. Type anything in the app name input
2. Clear the app name input
3. Press "Show me how to upgrade!"
4. You will see that the path of each file is wrong.

## Checklist

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -98,16 +98,13 @@ const SettingsContainer = styled.div`
   flex: 1;
 `
 
-const getAppInfoInURL = (): {
-  appPackage: string
-  appName: string
-} => {
+const getAppInfoInURL = () => {
   // Parses `/?name=RnDiffApp&package=com.rndiffapp` from URL
   const { name, package: pkg } = queryString.parse(window.location.search)
 
   return {
     appPackage: pkg as string,
-    appName: name as string,
+    appName: name as string | null,
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/components/pages/Home.tsx
```diff
@@ -143,11 +140,11 @@ const Home = () => {
...
 
   const appInfoInURL = getAppInfoInURL()
-  const [appName, setAppName] = useState<string>(appInfoInURL.appName)
-  const [appPackage, setAppPackage] = useState<string>(appInfoInURL.appPackage)
+  const [appName, setAppName] = useState(appInfoInURL.appName)
+  const [appPackage, setAppPackage] = useState(appInfoInURL.appPackage)
 
   // Avoid UI lag when typing.
-  const deferredAppName = useDeferredValue(appName)
+  const deferredAppName = useDeferredValue(appName || DEFAULT_APP_NAME)
```
pvinis, line 147:
```suggestion
  const deferredAppName = useDeferredValue(appName ?? DEFAULT_APP_NAME)
```

src/components/pages/Home.tsx
```diff
@@ -281,7 +278,7 @@ const Home = () => {
                 <Input
                   size="large"
                   placeholder={DEFAULT_APP_NAME}
-                  value={appName}
+                  value={appName || ''}
```
pvinis, line 281:
```suggestion
                  value={appName ?? ''}
```

pvinis (approved):
could you apply these to make things more typescript-y, and we can merge. thank you for this fix!

vinzscam:
> could you apply these to make things more typescript-y, and we can merge. thank you for this fix!

I went with inferred types here to avoid redundant typing since TypeScript handles that really well. But if there’s a spot where you think explicit types would help, let me know, and I can adjust!

src/components/pages/Home.tsx
pvinis: @vinzscam you can just press the button here to apply these.

src/components/pages/Home.tsx
vinzscam: this would be actually wrong as I want to use `DEFAULT_APP_NAME` if `appName` is empty

pvinis: hm.. i see. alright.

pvinis (approved):

PR #397, opened 2025-04-04
Add warning for AppDelegate.mm on RN 0.77

thiagobrez:
# Summary

This PR adds a warning to the "Useful Contents" section when upgrading to RN 0.77 or above.

The issue we're trying to warn is because on 0.77, the AppDelegate template was migrated from Obj-C++ to Swift, but it is not only a syntax change. There were more lines added that, if the user decides to stick with the `AppDelegate.mm` version - and I imagine most people with existing apps will do - will pass unnoticed, since the diff just shows that the Obj-C++ file was deleted.

The message calls this out, and points to the blog post where the core team highlight the needed change.

Posted about it here: https://x.com/thiagobrez/status/1908101612838572473

CleanShot 2025-04-04 at 17 06 03

## Test Plan

Going from any version below 0.77.0, to any version above 0.77.0, should show this message.

Select FROM 0.76.9 TO 0.77.0, or any other combination as explained above.

## Checklist

### DIFF
--- a/src/releases/index.js
+++ b/src/releases/index.js
@@ -2,6 +2,7 @@ import { PACKAGE_NAMES } from '../constants'
 
 const versionsWithContent = {
   [PACKAGE_NAMES.RN]: [
+    '0.77',
     '0.73',
     '0.74',
     '0.72',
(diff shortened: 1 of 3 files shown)


thiagobrez:
cc @diegolmello

pavlos-sqds (approved):

pvinis (approved):

PR #404, opened 2025-06-14
Update AlignDepsAlert.tsx

pvinis:
# Summary

## Checklist


lucasbento (approved):

.node-version
```diff
@@ -1 +0,0 @@
-16.19.1
```
kelset, line 1: should we have something equivalent (but recent) in the package.json? the engine field? I think we realistically dont want to be under Node 22 (but i might be wrong)

.github/workflows/push.yml
```diff
@@ -32,23 +32,23 @@ jobs:
...
-        with:
-          name: ${{ github.sha }}
+  # lint:
+  #   needs: [setup, install]
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - name: ⬇️ Restore output
+  #       uses: actions/download-artifact@v4
+  #       with:
+  #         name: ${{ github.sha }}
```
kelset, line 42: did you just commented them out temporarely or it's intentional to have it commented here?

.node-version
pvinis: its part of the mise.toml file :) ill update everywhere else too.

.github/workflows/push.yml
pvinis: i wanted to see if i can just get this running or not. not a permanent thing!

kelset:
Thanks for this, Pavlos — I’m doing a maintenance pass on the open PRs and finally taking a pass through some of the older ones.

I think the useful part of this PR is the extra `.xcodeproj/project.pbxproj` guidance in the alert copy. But at this point the branch is old enough that it also includes a few stale repo-level changes that I don’t think we should land as part of the same PR:

- the lint/typecheck job changes in `.github/workflows/push.yml`
- removal of `.node-version`
- incidental lockfile churn

So I don’t think this is something we should merge as-is anymore.

I’m going to close this one out rather than try to salvage the mixed branch. If we still want the xcodeproj guidance, I think the better move would be to reintroduce that part separately in a fresh, narrowly scoped PR from current `master`.

PR #409, opened 2025-07-10
feat: set areAllCollapsed as true in done section (fixes #403)

### LINKED ISSUE #403: When a file is marked as done, it should collapse too.
# Feature Request

Right now when marking a file as done, it goes to the bottom part with the `Done` files. It should also collapse, so it takes less space.

adrielfabricio:
# Summary

This pull request Resolves #403.

It implements **automatic collapse** for files marked as completed. The logic was implemented by modifying the `useState` hook to check whether the current section is `Done`; if so, the `areAllCollapsed` variable is set to `true` by default.

This change improves the user experience by reducing visual clutter in the "Done" section.

## Steps to Reproduce

1. Add a comparison between two versions of React Native.
2. Mark any file as completed.
3. Go to the **Done** section and confirm that the completed file is automatically collapsed.

## Checklist

- [x] I added documentation in `README.md` (if needed)


pvinis:
waiting for the preview to test, but I don't think this resolves the issue. when I mark the first individual file as done, it should move down and also collapse.

pvinis (approved):
my bad. it totally makes sense. thanks for the PR!

kundan175 (approved):

kundan175 (approved):

kelset (approved):
Thanks for the PR — I'm doing a maintenance pass on the repo and finally had a chance to review this one properly.

I think this is a nice small fix and I'm comfortable with it landing. The change is narrow, matches the issue well, and the approach makes sense for how the done section is rendered today.

I also ran the usual local checks on the branch:
- `yarn lint`
- `yarn typecheck`
- `yarn test --runInBand`

From my side this looks good to merge.

